PR #122, opened 2015-01-04
New string expander

JensAyton:
Original blurb from mailing list:
> The goal is to remove the use of printf-style formatting operators in descriptions.plist (for both safety and consistency with the JS way of doing things), and also simplify call sites where string formatting is used. The downside is that the implementation requires some fairly deep macro magic.

> The basic idea is that instead of code like this:
> ```objc
> [NSString stringWithFormat:DESC(@"format-foo-f"), foo]
> // format-foo-f expands to "bla bla %.1f"
> ```
or:
> ```objc
> OOExpandDescriptionString(@"[format-bar]", [UNIVERSE systemSeed], [NSDictionary dictionaryWithObject:[NSNumber numberWithInteger:bar] forKey:@"bar"], nil, nil, 0)
> // format-bar expands to "bla bla [bar]"
> ```

> you will get:
> ```objc
> OOExpand(@"[format-baz]", baz)
> // format-baz expands to "bla bla [baz|precision:1]"
> ```

> As per the example, this also adds a concept of formatting operators, to handle things like numerical precision. Other operators so far are "cr" and "dcr" for credit formatting (dcr takes a value in tenths of credits).

To do:
- [x] Test on Windows
- [x] Test on Linux
- [x] Document changed keys for translators

To don’t:
- <del>[ ] Convert every single thing to the new way before merging</del>


cim--:
Seems to work here

PR #126, opened 2015-03-09
Added ability to increase and decrease the field of view.

DaymareOn:
This code is for review. Even if it works, it would benefit from some beautifying: renaming of variables, methods,  moving methods to the right place in the class, and of course I take every advice :-)


AnotherCommander:
Looking pretty good overall. Thanks, good job,

Little bugs found: FOV is applied everywhere, even where it normally shouldn't like:
- Shipyard screen
- Planet info (F7) screen. You can actually change the FOV realtime with the effect of zooming the miniature planet in/out in that scree.
- If there are backgrounds in the GUI screens, their FOV changes as well when in flight. Interesting effect, but probably not desired. Check what happens with BGS OXP installed.
- If the player ship is set to be shown in the Commander Status screen, its FOV changes realtime when in flight.

Regarding the feature itself, I feel that it is more suitable as a game setting rather than a gameplay mechanic. Having it settable via the Game Options screen would be great. Also, if it can be made accessible to JS, then OXPs could set it and this could also mean usage by JS frame callbacks, so equipment based on this can be made etc.

I imagine this feature would probably make a very cool effect at the end of the hyperspace contdown (spatial distortion effect etc). I see a lot of potential in this.

DaymareOn:
All bugs aknowledged :-)
Ok with suggestions.

I thought to have a wave distortion effect when a laser hit our ship with energy level empty :-p

cim--:
> I feel that it is more suitable as a game setting rather than a gameplay mechanic. Having it settable via the Game Options screen would be great. Also, if it can be made accessible to JS, then OXPs could set it and this could also mean usage by JS frame callbacks

I think it should probably either be a Game Options setting - to calibrate it to display hardware appropriately - _or_ an in-game setting changeable by JS. I suppose you could set the default FOV in Game Options, and then have the in-game setting be a relative FOV which scales around 1.0?

DaymareOn:
Hi cim,

default FOV in game options, then relative in-game settings, seems to be a perfectly workable idea to me.

"display hardware": are there some different hardwares which would benefit from a calibrated field of view?

I would have thought that the default FOV could be set by the ship model. I'd imagine perfectly an andromeda with a fixed (or not very variable) fisheye. (And yes I'm thinking of a Dune Guild Navigator).
But this kind of oxp'ing makes sense mainly if the standard FOV is usable.

cim--:
> are there some different hardwares which would benefit from a calibrated field of view?

There was someone with three monitors a while back who wanted a wider FOV so that the image stretched properly around, I think.

AnotherCommander:
Guys, OK to bring this one in?

My only objection is the assignment of keys. After playing with it a bit, I am not very happy to have FOV changing during gameplay.I've noticed that with some ship models, big FOV changes can bring into the field of view the model itself and that just looks wrong.Also, I don't like having two new keys assigned to the same feature. I would love it though as game option and, since DaymareOn would like to have it, my proposal is this: Merge in this request, then use a compile-time switch to enable or disable the assignment of keys and in-combat changing of FOV. We can even let the switch set to enable for a few nightlies, to see what the general perception of the testers is.

Looking forward to your opinions.

DaymareOn:
I find the compile switch a very good idea.

It may even be temporary if we could assign keys through oxp.
Something like: an oxp may subscribe a functionality to the list of "functionalities to which one may assign keys/buttons/axis", and then the player may assign as he sees fit.
This way, the number of keys for the core game is not modified, and players may do as they want.

AnotherCommander:
@DaymareOn : We jhave a;ready something like that in place, by means of prime-able equipment activation  key assignments. By default the game uses it for immediate activation of cloaking device but any prime-able equipment can be used instead. It is not an ideal solution, but it is a possibility.

DaymareOn:
Prime-able equipment activation key assignment: wow, it's a good news that it's already implemented.

JS read/write access of the FOV properties: yes, of course. It's in my Issues list :-)
(I first want to address all known bugs in the Issues list, which in turn required me to use Doxygen.)

KonstantinosSykas:
AnotherCommander commented: 
"....I've noticed that with some ship models, big FOV changes can bring into the field of view the model itself and that just looks wrong..."

I think this goes toward introducing a cockpit.

I also like DaymareOn's idea of having the default FOV set by the ship model.

DaymareOn:
Viewing the model at big FOVs: well... I really liked to be able to see the Cobra mk3 laser cannon during the tests, and even the reactors on the rear view. 

I think that if the max FOV is set by the ship oxp, it may be a functionality rather than a bug. 

Yet I'm ok that this needs to be prevented by default so as not to break the player's immersion. 

Maybe this will converge towards a ShipMaxFieldOfView different from the EquipmentMaxFieldOfView, which would be specific to the equipment enabling the FOV movement. (Ie "sniper lens equipment", "universe consciousness equipment" (fixed fisheye), etc.).

cim--:
One possibility on the keys is that we don't necessarily have to assign default values for everything. Most players probably have a keyboard with a numpad, and/or separate F-keys, which gives them a fair bit of extra space to assign functions to. We could have `key_fov_increase = ""; key_fov_decrease = "";` in the keyconfig.plist, and let people set them if they want. (Same for other potential new functions which don't necessarily need their own key for everyone)

I do think we should either allow assigning keys or make it a Game Options setting, but not both. For JS and ship OXP settings, probably this tends towards it not being Game Options, for me.

DaymareOn:
I think fov manipulation by key shouldn't be available by default.
It should be a functionality enabled by oxps with a suitable handwavium: fisheye zoom for traders, telephoto zoom for sightseers, specially crafted zoom for fighters, etc.

With the Game Options menu, you could set the default FOV (which is calibrated for the default ship), then the initial_fov is the default fov multiplied by a factor depending on the ship, then depending on the oxp, you may use the keys to vary your fov.

But it means that the keys usage must be enabled in this use case.

AnotherCommander:
For me, the best way this would work in-game is the exact same way Gamma works: A game setting. The way I see it, it has the same level of functionality. It is just a preference. Some people want to see things wide, while others may enjoy a different sense of scale by closing somewhat the field of view. Other than that, I could accept a read/write JS setting for it and that's about it. No special key assignments or anything like that. I do believe it is a great feature to have, but I would not go as far as making such functionality a gameplay mechanic.

I did propose the compile-switch option for in-game controls because I understand that there could be people who might find it interesting and doing it this way may allow us to make a decision for enabling / disabling, while at the same time we can pull in this request. My personal preference would be for this switch to be default off.

 I am on vacation at the moment and away from my dev computer, so I don't think I will be able to act on this request in any case until the first week of April. I will try to follow the discussion as much as I can though and hopefully we'll have reached a conclusion on how exactly we want this incorporated by the time I'm back. However, anyone with merge rights for external requests, please feel free  to act on it if you want in case a consesus has been reached before my return.

AnotherCommander:
Unless someone has a good reason not to, I am planning to merge this request either tonight or sometime tomorrow. There is still some work to be done, but I believe that right now this is at a good starting point.

DaymareOn:
I've fixed all the bugs I was aware of. I'm currently working on having FOV settable via the Game Options screen.

AnotherCommander:
I think I found a little bug with the implementation. Look at this screenie:
image
This is with FOV set to 0.75 (which I believe, should correspond to 43 degrees (the fieldOfView variable is expressed in radians if I understand the implementation correctly). I am targeting a station, but the targeting square brackets are not closing around the station properly when at the edges of the screen. They also seem to move at noticeably different speed to the station when I turn the ship around. It is quite a confusing and weird effect. Maybe an adjustment can be made so that it is not as jarring as it stands now?

DaymareOn:
Wow! Thank you :-) I'll look into the targeting square brackets issue.

The vertical field of view angle is 2 * tan-1(field of view variable).
The horizontal field of view angle depends on your screen.

The easy way to apprehend this is to think that your screen is a rectangle set at 1 unity away from you, with a height of 2 * field of view variable (so 1 by default) unities.
I would be ok to use a different variable (angle in degrees for example) to have it more easily readable.
I plan to have this in the javascript methods, at least.

AnotherCommander:
I don't want to go off-topic too much, but I'd just like to understand because I think I see differently in practice. Here are three (rescaled for convenience) screens of the same scene, taken in the same game session. They were all taken right after launch, with planet ahead and station behind, with the fieldOfView value set to 3.14159, which in rad is M_PI and should correspond to 180 degrees. One is very wide, one is very tall and one is the rescaled 1920x0180 fullscreen image. As you can see, the horizontal FOV (unless I have confused the definitions of horizontal and vertical) seems to show correctly a 180 degrees view, without any variation horizontally..
fov180vwide
fov18016_9
fov180vtall
:Now, the vertical FOV seems to change between images but that does not seem to agree with what you mentioned above.

It is very possible that I am mixing up horizontal and vertical, but in any case, from what I can understand from the posted images, fieldOfView seems to be indeed a rad value, which would be very convenient for quick conversion to degrees.

On a spearate note, please let me pull the request as it is now before going ahead to add scripting stuff.  I amy also be able to work on the option thing as soon as this request has been imported.

DaymareOn:
Well, in Universe.m, around line 4401, we do a OOGLFrustum(-ratio, ratio, -aspect*ratio, aspect*ratio, nearPlane, farPlane), with ratio equals to fov.

So, maybe you can check what you understand according to the definition of OOGLFrustum :-)

--- a/src/Core/Universe.m
+++ b/src/Core/Universe.m
@@ -4401,7 +4403,8 @@ static const OOMatrix     starboard_matrix =
                        {
                                float   nearPlane = vdist ? 1.0 : INTERMEDIATE_CLEAR_DEPTH;
                                float   farPlane = vdist ? INTERMEDIATE_CLEAR_DEPTH : MAX_CLEAR_DEPTH;
                               float   ratio = (displayGUI ? 0.5 : fov) * nearPlane; // 0.5 is field of view ratio for GUIs
                               OOLog(@"FOV",@"Universe ratio: %f", ratio);
                                
                                OOGLResetProjection();
                                OOGLFrustum(-ratio, ratio, -aspect*ratio, aspect*ratio, nearPlane, farPlane);

Regarding the pull request, well if you're ok to import the targeting bracket issue, it's totally ok with me :-)

kanthoney:
The OOGLFrustum arguments are left, right, bottom, top, near and far in that order.  So it looks to me like you're fixing the horizontal FOV and then calculating the vertical one from that, which is what it looks like from the screenies.

DaymareOn:
@kanthoney: You're right, I don't know why I remembered falsely that it was the vertical fov that was fixed :-/
Ok, so it's the horizontal plane that we work on, and the vertical one depends on the aspect ratio.

@AnotherCommander: Concerning the angle of view, I think that the tangent of the horizontal half-angle is the fov variable divided by the nearPlane.
tan(alpha/2) = fov / 1
=> alpha = 2 * atan(fov)
(I think of it as a triangle rectangle: the near side is the distance to the screen, the opposed side is the half width of the screen, so the tangent equals the opposed side divided by the near side)

If you use a fov of Pi, it means you're 1 unity far away from the screen, and the screen is 2*Pi wide.
So, your angle should be : 2 * atan(Pi) = 2.53 rad

From what I understand, it means you can never reach a 180° angle, because as we are one unity far from the screen, we would need a screen with an infinite width to reach 180°.

AnotherCommander:
Thanks all for your explanations. I just wanted to be sure how to calculate the fov angle in degrees because that's how I am planning to have it shown in the Options.

@DaymareOn : I will import the request as it is at this moment, but by all means, please continue on the fix for the targeting brackets.

PR #129, opened 2015-04-11
Target reticle positionning now takes the field of view into account.

DaymareOn:
This fixes the target reticle positionning issue in all 4 views + custom.

### DIFF
--- a/src/Core/HeadUpDisplay.m
+++ b/src/Core/HeadUpDisplay.m
@@ -3431,7 +3431,7 @@ static void hudDrawReticleOnTarget(Entity *target, PlayerEntity *player1, GLfloa
 	p1 = HPVectorToVector(HPvector_subtract([target position], [player1 viewpointPosition]));
 	
 	GLfloat			rdist = magnitude(p1);
-	GLfloat			rsize = [target collisionRadius];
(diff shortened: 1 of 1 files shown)


DaymareOn:
I must thank you for the code review you did on the #126 merge, and for the slider in game options :-)

Concerning fov scale, fov in radians, and fov in degrees, I think there is a little misunderstanding.
The screen window is at 1 unity of distance from the camera, and is one unity wide.
What we do with the fov functionality is to change the width (and height) of the screen window.
The _fov variable is really the half screen width seen from a distance of one unity.
So, the angle is 2 * atan(_fov), and it is in degrees or radians depending on the used atan function.

AnotherCommander:
Hmm... I guess renaming 
- (void) setFov:(float)value fromRadians:(BOOL)fromRadians
to
- (void) setFov:(float)value fromFraction:(BOOL)fromFraction and 

- (float) fov:(BOOL)inRadians
to
- (float) fov:(BOOL)inFraction

would help make things clearer for anyone reading the code?

DaymareOn:
Totally :-)

One thing I plan to do is to change this exposed standard value of 0.5 to an exposed standard value of 1.0, to:
- reduce the occurence of magic numbers,
- have a more obvious meaning of fov for oxp makers.

PR #137, opened 2015-05-08
Attempt to fix scissor test failure (issue 136)

kanthoney:
This is an attempt to fix issue 136.  There are two possible fixes, selected by setting the `issue_136_fix` user default to `0` (off), `1` (first fix) or `2` (second fix).

As I can't test whether the code builds, I'd appreciate somebody else at least looking over it before merging to master.

### DIFF
--- a/src/Cocoa/MyOpenGLView.m
+++ b/src/Cocoa/MyOpenGLView.m
@@ -324,7 +324,7 @@ static void UnapplyCursorState(OOMouseInteractionMode mode);
 
 - (void) initialiseGLWithSize:(NSSize)v_size
 {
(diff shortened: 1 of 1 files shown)


AnotherCommander:
If you don't get a response by someone who can actually test, best if you merge it anyway. We will see what happens with tomorrow's nightly on terrastorage; if there is a problem, the exact error message should be shown in the build diagnostics section.

cim--:
Modern git(hub) makes it very easy to unmerge code from pull requests, so agreed, merge it anyway and we'll either get a build result or can hand it to a Mac player for actual testing.

PR #140, opened 2015-05-09
Fast enumeration migration

MaddTheSane:
These patches migrate the usage of many `NSEnumeration` to this project's `foreach` and `foreachkey`, which translates to `for(...in...)` on systems that support it.

### DIFF
--- a/src/Core/AI.m
+++ b/src/Core/AI.m
@@ -803,7 +803,6 @@ static AIStackElement *sStack = NULL;
 	NSDictionary			*newSM = nil;
 	NSMutableDictionary		*cleanSM = nil;
 	OOCacheManager			*cacheMgr = [OOCacheManager sharedCache];
-	NSEnumerator			*stateEnum = nil;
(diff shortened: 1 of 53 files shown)


kanthoney:
Here's what the first screen after game start looks like under Linux:

snapshot3

MaddTheSane:
You got farther than I did when I tried to launch it (although I do have my own patches…). I'm also discovering places where the code expects values in an `NSDictionary` and not the keys.

I'll be fixing some of them soon.

Later: Okay, now I'm getting to the point with the blank screen.

MaddTheSane:
oops, one moment.

MaddTheSane:
I accidentally pushed a commit that changed the submodules. This should be fixed now.

kanthoney:
So you're issuing a pull request without actually getting it to work first?

MaddTheSane:
I assumed that the edits wouldn't break the game. I was wrong.

PR #141, opened 2015-05-10
Enum migration

MaddTheSane:
These patches migrate the usage of many `NSEnumeration` to this project's `foreach` and `foreachkey`, which translates to `for(...in...)` on systems that support it.

See also #140

### DIFF
--- a/src/Core/AI.m
+++ b/src/Core/AI.m
@@ -803,7 +803,6 @@ static AIStackElement *sStack = NULL;
 	NSDictionary			*newSM = nil;
 	NSMutableDictionary		*cleanSM = nil;
 	OOCacheManager			*cacheMgr = [OOCacheManager sharedCache];
-	NSEnumerator			*stateEnum = nil;
 	NSString				*stateKey = nil;
 	NSDictionary			*stateHandlers = nil;
 	NSAutoreleasePool		*pool = nil;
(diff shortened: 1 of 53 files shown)


mcarans:
I had another look at this and now there are tests, it may be worth looking into as it does not seem like a complicated change. Basically this needs to be brought up to date with master so we can see if the tests pass @oocube

MaddTheSane:
rebased.

mcarans:
@MaddTheSane It looks like there are some compile issues. Would you be able to look into those?

MaddTheSane:
I've tried to compile it using Xcode but encountered unrelated errors.

mcarans:
@MaddTheSane Thanks for contributing this improvement. I tested locally and it seems fine. I hope to get the Mac build working again in future probably by generating an XCode project from meson or cmake. 

I added the code below to make fast enumeration the default (ie. no need for the #define and #if lines to check for it). I don't have rights to add it to your fork. Please can you add this:

[0001-Always-use-fast-enumeration.patch](https://github.com/user-attachments/files/26641306/0001-Always-use-fast-enumeration.patch)

MaddTheSane:
Perhaps those can be a separate patch? You can also open up a pull request on my fork and branch.

mcarans:
@MaddTheSane I've merged your changes and added a separate PR for making fast enumeration the only option https://github.com/OoliteProject/oolite/pull/598

Issue #80, opened 2014-01-27
Player Ship model sown on F5 screen needs to be positioned better on the screen


BlueSteelAUS:
with the new ship display on F5 screen in VERSION_STRING=1.79.0.5570-140128-530e1cf
At the moment the ship is like its being top/right aligned. needs to be positioned like the ones on the ships market screen

kanthoney:
Fixed by #143, and confirmed on Mac by Konstantinos.

PR #143, opened 2015-05-13
Gui planets

kanthoney:
Fix for issue mentioned on the mailing list where the planet in the system info screen was being rendered at unusual sizes when the window has odd aspect ratios.  Obviously, this can be delayed until after 1.82 release.

### DIFF
--- a/src/Core/Entities/PlayerEntity.m
+++ b/src/Core/Entities/PlayerEntity.m
@@ -3645,17 +3645,6 @@ NSComparisonResult marketSorterByMassUnit(id a, id b, void *market);
 	if (shipData == nil)  shipData = [[OOShipRegistry sharedRegistry] shipInfoForKey:shipKey];
 	if (shipData == nil)  return;
 	
-	// Correct the position of the displayed model in the status screen when viewed in non-4/3 resolutions. Other screens displaying
-	// demoship models remain unaltered - Nikos 20140129
-	NSSize screenSize = [[UNIVERSE gameView] viewSize];
-	GLfloat screenSizeCorrectionFactor = 1.0f;
-	if ([context isEqualToString:@"GUI_SCREEN_STATUS"])
-	{
-		screenSizeCorrectionFactor = screenSize.height <= screenSize.width ?
-										screenSize.height / screenSize.width * (4.0f/3.0f) :
-										screenSize.height / screenSize.width * (3.0f/4.0f);
-	}
-	
 	Quaternion		q2 = { (GLfloat)M_SQRT1_2, (GLfloat)M_SQRT1_2, (GLfloat)0.0f, (GLfloat)0.0f };
 	// MKW - retrieve last demo ships' orientation and release it
 	if( demoShip != nil )
@@ -3673,7 +3662,7 @@ NSComparisonResult marketSorterByMassUnit(id a, id b, void *market);
 	
 	GLfloat cr = [ship collisionRadius];
 	[ship setOrientation: q2];
-	[ship setPositionX:factorX * cr * screenSizeCorrectionFactor y:factorY * cr * screenSizeCorrectionFactor z:factorZ * cr];
+	[ship setPositionX:factorX * cr y:factorY * cr z:factorZ * cr];
 	[ship setScanClass: CLASS_NO_DRAW];
 	[ship setRoll: M_PI/10.0];
 	[ship setPitch: M_PI/25.0];
--- a/src/Core/Universe.m
+++ b/src/Core/Universe.m
@@ -4414,7 +4414,7 @@ static const OOMatrix	starboard_matrix =
 				}
 				else
 				{
-					OOGLFrustum(-ratio/aspect, ratio/aspect, -ratio, ratio, nearPlane, farPlane);
+					OOGLFrustum(-3*ratio/aspect/4, 3*ratio/aspect/4, -3*ratio/4, 3*ratio/4, nearPlane, farPlane);
 				}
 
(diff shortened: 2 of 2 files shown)


AnotherCommander:
I initially thought that this would be a very complex fix, but after having seen the results in the gui_planets branch, I feel that it works quite well and would make sense to pull it in before 1.82. Probably a good plan of action would be to have the branch tested also by Linux/Mac and if we are all happy that it works without side effects then bring it in. Would that be OK?

kanthoney:
There's a similar issue with very tall windows out in space (that's a gun - it's not that I'm very pleased to see someone!):

oolite-006

The fix should be similar, so I'll look at that.

kanthoney:
That's better:

oolite-008

cim--:
Looks good here. Combined with the new FOV setting this should make it a lot easier to get the right things on screen. Agree it should be merged before 1.82 assuming it tests okay.

AnotherCommander:
Tested on two systems (one with dual monitor setup) and it looks good here. Would be nice to confirm also on Mac, but I don't think that it would be any different there.

KonstantinosSykas:
Managed to do some testing on the Mac, and I saw some interesting things.
1. The window height and width sizes have a low limit so you cannot go to
extremely small sizes.
2. Start screen,
3. While in flight I don't know if we have a problem. See screenshots below
(same view different mode) and judge for yourselves.
(sorry for the bad graphics, but it was the only way to be able to have VNC
screen with a decent refresh rate)
Windowed mode (laser is showing):
http://s1124.photobucket.com/user/ksykas/media/Oolite%20Bugs/Oolite-ResRatio-Window-Mac.png.html
Full screen mode (laser is not showing):
http://s1124.photobucket.com/user/ksykas/media/Oolite%20Bugs/Oolite-ResRatio-FullScr-Mac.png.html

kanthoney:
For #3, on a tall screen the gun should look the same as it does for a
window with 4/3 aspect ratio - if not there's a problem somewhere.  (Stupid
question - are you up to date with that branch?)

Do all the graphics in the GUI work OK?

On Fri, May 15, 2015 at 3:34 PM, Konstantinos Sykas <
notifications@github.com> wrote:

> Managed to do some testing on the Mac, and I saw some interesting things.
> 1. The window height and width sizes have a low limit so you cannot go to
> extremely small sizes.
> 2. Start screen,
> 3. While in flight I don't know if we have a problem. See screenshots below
> (same view different mode) and judge for yourselves.
> (sorry for the bad graphics, but it was the only way to be able to have VNC
> screen with a decent refresh rate)
> Windowed mode (laser is showing):
>
> http://s1124.photobucket.com/user/ksykas/media/Oolite%20Bugs/Oolite-ResRatio-Window-Mac.png.html
> Full screen mode (laser is not showing):
>
> http://s1124.photobucket.com/user/ksykas/media/Oolite%20Bugs/Oolite-ResRatio-FullScr-Mac.png.html
>
>
>
>
> On Thu, May 14, 2015 at 8:01 PM, cim <notifications@github.com> wrote:
>
> > Looks good here. Combined with the new FOV setting this should make it a
> > lot easier to get the right things on screen. Agree it should be merged
> > before 1.82 assuming it tests okay.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/OoliteProject/oolite/pull/143#issuecomment-102103267
> >.
>
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/OoliteProject/oolite/pull/143#issuecomment-102415141>.
>



-- 
Kevin Anthoney
kevin@kanthoney.co.uk

KonstantinosSykas:
Start screen, start new commander screen, ships for sale screen and data on planet screen are OK
However, please keep in mind that I cannot check extreme small sizes.

Concerning your latest comment: Indeed the gun is showing in 4:3 ratio. However when I start widening the window, the moment I pass from 1:1 ratio then the gun starts hiding.

KonstantinosSykas:
Let me say this correctly this time because the previous explanation was wrong.
Starting at 4:3 the weapon is showing. Widening the window the weapon starts hiding.

kanthoney:
I guess we can merge up to https://github.com/OoliteProject/oolite/commit/e83ce6acfd86d8142ade6ed15a9ee98fb882f371 anyway.    I'll see if I can find anything obvious about the problem with the gun.

kanthoney:
I've just seen your last message - that sounds like correct behaviour to me, except I'm not sure why you're seeing the gun at 4:3.  Have you tweaked the FOV settings?

AnotherCommander:
Might as well merge it all and try to fix Mac in the meantime. I don't see why we should leave out a fix that already works for two out of three platformst.

AnotherCommander:
Actually it seems to work as intended. Here are screenies in various resolutions and FOV of  
oolite-001
oolite-002
oolite-003

KonstantinosSykas:
No FOV change. I see it at the Options menu (I don't know where Mac is saving the defaults!) and has a value of....what???....arghhh...65??? 65??? Where did that come from?

OK. Setting this at 57 and no gun is showing at 4:3.

Question: How come the default FOV for Mac is 65? I am pretty sure that I have not touched that anyhow.

KonstantinosSykas:
Just let it be. With all these changes that I've tested it could be that I also changed the FOV. I think we are OK with the fix.

KonstantinosSykas:
This ends the paranoia. Thank you Kevin. :-)

cim--:
Has this also fixed #80?

Issue #139, opened 2015-05-09
doesn't compile


lee-:
compiling on gentoo:

 Compiling file src/Core/Entities/SkyEntity.m ...
In file included from src/Core/MiniZip/ioapi.c:24:0:
src/Core/MiniZip/ioapi.h:127:35: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘OF’
 typedef voidpf   (ZCALLBACK *open_file_func)      OF((voidpf opaque, const char* filename, int mode));
                                   ^
src/Core/MiniZip/ioapi.h:128:34: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘OF’
 typedef uLong    (ZCALLBACK *read_file_func)      OF((voidpf opaque, voidpf stream, void* buf, uLong size));
                                  ^
src/Core/MiniZip/ioapi.h:129:35: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘OF’
 typedef uLong    (ZCALLBACK *write_file_func)     OF((voidpf opaque, voidpf stream, const void* buf, uLong size));
                                   ^
src/Core/MiniZip/ioapi.h:130:33: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘OF’
 typedef int      (ZCALLBACK *close_file_func)     OF((voidpf opaque, voidpf stream));
                                 ^
src/Core/MiniZip/ioapi.h:131:37: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘OF’
 typedef int      (ZCALLBACK *testerror_file_func) OF((voidpf opaque, voidpf stream));
                                     ^
src/Core/MiniZip/ioapi.h:133:33: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘OF’
 typedef long     (ZCALLBACK *tell_file_func)      OF((voidpf opaque, voidpf stream));
                                 ^
src/Core/MiniZip/ioapi.h:134:33: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘OF’
 typedef long     (ZCALLBACK *seek_file_func)      OF((voidpf opaque, voidpf stream, uLong offset, int origin));

... and so on

[...]

PR #144, opened 2015-05-13
Fix compilation on Gentoo

kanthoney:
Fix for issue 139, the compilation failure on Gentoo.  Also tested on Debian.


KonstantinosSykas:
I need some time to check making of Linux distro-neutral installer.

kanthoney:
No problem.  I don't see any point in putting this in pre-1.82 release anyway, as it's only going to affect people building from source.

KonstantinosSykas:
Making of Linux distro-neutral installer checked OK for both 32bit and 64bit architectures. 
May the Learned Elders of Oolite decide the future of this pull request.

cim--:
I think it should probably be merged - as I said on #139, a better post-1.82 fix would be to take the OF macro out of the Minizip code entirely. If it seems to work on all the various platforms we have it should be safe.

PR #146, opened 2015-05-20
Tumbling ships

kanthoney:
(I was going to leave this until post 1.82 release, but as it brought up a graphical glitch - and a fix for said glitch - I thought I'd let you look at it before release.)

I've often thought that the way the spinning ships in Oolite rotated was rather boring compared to the original Elite, where they tumbled in a much more interesting way.  So I thought I'd try and reproduce the Elite style rotation.

As it happened, this turned out to be an excellent stress test for the graphics engine, and it threw up a problem (see https://github.com/OoliteProject/oolite/commit/643566280a5f47c7455660ab2bc400f06d59438e).  After staring at this for a while, it looked to me as though any face which was exactly perpendicular to the x-axis wasn't being lit properly - it was either fully lit or fully unlit.  I think I've fixed this in https://github.com/OoliteProject/oolite/commit/e0cf14146b132a8b672b6501dd844d900f43cd7b, but there may be knock-on effects.


AnotherCommander:
It is the knock-on effects that I am worried most about. At this stage, we do not really have time to catcch such potential effects - we could maybe catch some but we cannot be certain that we got them all. Since the issue is not a major or game-breaking one, I would prefer leaving it out for now.

What do you guys think?

AnotherCommander:
I think I can still see the problem targeted by commit e0cf141. It is especially evident at the thruster area of the Fer-de-Lance in the ship library screen.

kanthoney:
I'm fine with leaving it until post release - I just wanted to give you the option.

kanthoney:
Just to explain a bit more what the problem actually is, if you look at the `FACES` section of one of the model files (e.g. `oolite_cobra3.dat`), you'll see that the face normals (which are the second batch of three coordinates) are all set to zero.

In the `generateFaceTangents` function in `OOMesh.m`, the tangent is set to the cross product of the face normal and another vector by this line:

    face->tangent = cross_product(nA, tangent);	// Rotate 90 degrees. Done this way because I'm too lazy to grok the code above.

Since the normals are zero, the tangents are zero too.  The tangents are used in the function `getNormal: andTangent:`, where the vertex tangents are set by this line:

    	*outTangent = vector_normal_or_fallback(tangent_sum, kBasisXVector);

`tangent_sum` is zero, so `outTangent` is always set to `kBasisXVector`.  If the vertex normal also happens to be parallel to the x direction then OpenGL gets confused by having the normal and the tangent in the same direction and doesn't get the lighting right.  Since the comment after the cross product line indicates that it's a hack, the fix is to try and do it properly instead.

AnotherCommander:
Thanks for the detailed explanation Kevin. I am sure you understand that the only reason I am not saying "yeah go ahead and do it" is basically because I am afraid that we may have just missed something and that that something might come back to bite us within hours of release.

I did test the feature briefly though and the fix seems to work, Here is something that we can do to have the code in the released 1.82 and at the same time take as little risk as possible:
- Import to 1.82 only the specific part of this request dealing with the lighting fix.
- To avoid last minute nast surprises, and seeing that the fix is basically a change of a few lines, introduce a defaults file settting named e.g. "lighting-fix-tentative", default it to NO.
- IF the user sets it to YES, drive the code towards the lighting fix, if not set or if set to NO, keep the existing code path.

Would that be worth of consideration?

cim--:
I like the effect, and now that 1.82 has been out a while it's probably worth bringing it in. The rotation overall does feel considerably faster than the previous one, though - perhaps half or even a third of the speed would be better?

AnotherCommander:
Agree on the rotation slowdown. It is very nice, but way too fast in its current form.

AnotherCommander:
While we are at it, we should also switch the "lighting-fix-tentative" default to YES,

kanthoney:
You can tweak the speed (and overall effect) by fiddling with the 6 and 5 in lines 2277-2280 of `ShipEntity.m`.  You can also tweak the quaternions in the next two lines - just make sure they're normalised. Finding a combination that gives a nice, balanced rotation is a bit fiddly.

I think merging this branch will remove the `lighting-fix-tentative` option altogether and leave it permanently on.

AnotherCommander:
Looks fine I think. I'll try to merge it at around lunch break if I can.

PR #147, opened 2015-05-20
Add fix for lighting bug, enabled by lighting-fix-tentative user default

kanthoney:
This is just the lighting fix from #146, enabled by setting the lighting-fix-tentative user default to YES.


AnotherCommander:
I think I am happy to have this imported. Are we all ok with this?

cim--:
Looks good to me, though why it works isn't something I can understand.

PR #149, opened 2015-05-23
Update shipdata.plist

Norbylite:
Fix engine materials in oolite_template_viper, complement missing engine materials in oolite_template_viper-pursuit.


Norbylite:
I tested this on my machine using Gallery, Vipers now has blue glow within the engine tubes and Viper pursuits has textures on engines instead of the purple filling color.

Issue #157, opened 2015-08-07
build failures on Xcode 7 beta 5


MaddTheSane:
While investigating issue #152, I decided to build and test the app. First, I built it on Xcode 6.4, and it build fine (with a lot of warnings, mostly those with `-Wcstring-format-directive` (Having a C string specifier (`%s`) in an `NSString` or `CFStringRef` formatting string). I then tried to build it on Xcode 7. 

There's a few errors, most of them are due to more strict compiler checks. The leading problem is sending `nil` to functions/methods that require a non-null argument. There's also a couple of "designated initializer missing a 'super' call to a designated initializer of the super class" for `OOJSInterfaceDefinition` and `OOJSPopulatorDefinition`.

I'll see if I can't  create a pull request with fixes, as there's only 10 errors in the *issue_152* branch.

AnotherCommander:
Oolite now builds on Xcode 7. Issue closed.

PR #158, opened 2015-08-07
Fix the project to build under Xcode 7.

MaddTheSane:
Addresses #157.


cim--:
Thanks for the fixes - they all look fine. I'll confirm that they don't cause any problems on the GNUStep build and merge if so tomorrow.

PR #161, opened 2015-08-22
Multiple laser mounts

cim--:
This is a rewrite of the laser mount code to allow multiple mount points to be fitted on a single direction in a nicer way than the "subentity" pattern, including fitting multiple lasers on aft or side mounts, and including player ships.

`weapon_mount_mode` is a new shipdata property which can be "single", "split" or "multiply". "single"is the default value and duplicates 1.82 behaviour for backward compatibility. The other two change the expected value of `weapon_position_forward` (and the other three) from a vector expression to an array of vector expressions. "split" then splits the damage from the weapon across those positions - it's cosmetic only, so long as you hit with all of them; "multiply" is (mostly) equivalent to doing it with subentities in that each beam is a full power shot (though unlike the subentity solution, this has the heat and energy costs of that many full shots). You can't split some directions and multiply others, but I think that would be overcomplicating things - you can always make some custom lasers to get the same effect if you really want to do that.

Here's an experiment with a Krait
oolite-617
```
weapon_mount_mode = "split";
weapon_position_aft = ("0.0 0.0 -34.6");
weapon_position_forward = (
	"45 -1.5 7",
	"-45 -1.5 7"
);
weapon_position_port = ("-45.0 0.0 -13.5");
weapon_position_starboard = ("45.0 0.0 -13.5");
```

There is a minor incompatibility for OXPs - `ship.weaponPositionForward` is now an array of vectors rather than a vector - but that was only added in 1.77 and I don't think that much makes use of it.

(Something which still probably needs doing: if the player ship has "multiply" weapons, the price of lasers should be multiplied by the number of mounts)

### DIFF
--- a/src/Core/Entities/PlayerEntity.h
+++ b/src/Core/Entities/PlayerEntity.h
@@ -941,7 +941,7 @@ typedef enum
 
 - (OOWeaponType) weaponForFacing:(OOWeaponFacing)facing;
 - (OOWeaponType) currentWeapon;
-- (Vector) currentLaserOffset;
+- (NSArray *) currentLaserOffset;
 
 - (void) rotateCargo;
 
--- a/src/Core/Entities/PlayerEntity.m
+++ b/src/Core/Entities/PlayerEntity.m
@@ -8796,15 +8796,15 @@ static NSString *last_outfitting_key=nil;
 
 				}
 				
+				NSString *timeString = [UNIVERSE shortTimeDescription:installTime];
 				NSString *priceString = [NSString stringWithFormat:@" %@ ", OOCredits(price)];
 
-				NSString *timeString = [UNIVERSE shortTimeDescription:installTime];
-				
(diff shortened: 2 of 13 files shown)


Zireael07:
Wooooo!!!!

Can't wait to see it in-game!

AnotherCommander:
Looks and works good. I tried it with a three-mount-laser modified player Cobra and got in a fight with a Krait modified as above. All fne as far as I can see,

AnotherCommander:
Good to pull into master. Just one suggestion: can the multiply and/or split congifurations be made JS-controlled for potential in-scrpt switching?

cim--:
Technically easy enough, but it would allow a "free money" exploit: set the mode to "split" and buy a military laser for 6000, then set to "multiply" and sell it for 12,000 or more.

If you want a ship that can switch modes, then it can be worked around by making a custom laser with twice (3x, etc.) the energy usage, heat and damage of the standard one, and swapping that in for "multiply" mode.

PR #171, opened 2015-11-26
Adding a sell feature for equipment.

ocz1:
First things first. I hate git (the tool). This should be all files I have edited, but I struggled quite hard just to reach this point, where I can commit this proposal. On the way I lost the ability to compile the project to see if everything works. I did compile a running version before I tried to upload it to github and it worked quite well ( or I wouldn't have decided to upload it.)

This changes add a new possible key for an item to equipment.plist: "can_be_sold"
The new method in ShipEntity.m "canAddEquipmentForSale" checks if an item has this new key, if its value is "true" and if this item already has been installed.
Another check has been added to the equipment purchase GUI in PlayerEntity.m , that uses "canAddEquipmentForSale", to see, if it's installed and has the new key. In this case the item's identifier key will be left to be listed in the purchase screen, but later on its appearance changes. The textcolor of the entry changes to orange, much like the "repair option", the string "Sell: " is added to the title and other disclaimers are added to the description. Further the behaviour, when selected, changes to refunding the price and removing that item from the ship.

Pro:
- OXP developers get a simple way to add a sell option.
- This sell option provides a unified look for all Items that use it.
- The look of the sell option differs from regular purchase list entries, making them easier to spot and more distinguishable from other entries.

Contra:
- The sell option, as simple as it is, is quite inflexible at this point. If an OXP developer wants to do something different to sell an Item (different refund value, diff. techlevel, edited description texts, dependences to other equipment). In this case she/he has to fall back on using the old methods. Those methods don't provide the different "Sell Option" color.

There are two ways to get rid of that contra point.
- Adding more Keys to the item, that contain alternative data like techlevel or refund value. This may be implemented much like the "weapon_info" key. A "sell_info" key.
- The second way is kinda cheap, but maybe the better solution. Adding another key that let developers change the textcolor of the purchase entry. This way they can mimic the design of the sell option (and as bonus of the repair option too) and have all the same possibilities as before. Choosing the color wisely should be left in the devel.'s hands. (Screwing around with colors just shows bad taste of the devel. and of the OXP and not of the game as a whole.) The colors can be limited to the colors of the repair and sell options, too. "entry_color" = "blue"; (red, orange, aso. or standard_color, repair_color, sell_color).

In any case a sell option in any form should left in-code, to give an example how an entry for selling should look like. Ohh, and my changes aren't usable on items, that can be installed multiple times.

### DIFF
--- a/Resources/Config/descriptions.plist
+++ b/Resources/Config/descriptions.plist
@@ -1656,9 +1656,12 @@
 	"equip-cash-value"				= "Cash: [credits|dcr].";
 	"equip-no-equipment-available-for-purchase" = "No equipment available for purchase.";
 	"equip-repair-@"				= " Repair:%@";
+	"equip-sell-@"					= " Sell:%@";
+	"upgradeinfo-@-price-is-for-refunding" = "%@ (Price will be refunded for the existing system).";
 	"upgradeinfo-@-price-is-for-repairing" = "%@ (Price is for repairing the existing system).";
 	"@-will-replace-other-energy"	="%@ (The installed energy recharge unit will be removed and sold for scrap).";
 	"upgradeinfo-@-weight-d-of-equipment" = "%@\nTakes up %dt of cargo space.";
+	"upgradeinfo-@-frees-weight-d-of-equipment" = "%@\nFrees up %dt of cargo space.";
 	"forward-facing-string"			= " Forward ";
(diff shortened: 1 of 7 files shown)


ocz1:
Chances are, I take back my proposal, as I believe I found a better approach for offering sell options.

PR #172, opened 2015-11-30
Sell option.

ocz1:
Equipments with identifiers ending with the "_SELL" suffix get a special treatment as "sell options".
This different treatment includes:
- Different textcolor in the equipment purchase menu (similar to the already implemented "repair options")
- Price values are added to the player account instead of being subtracted.
- {Price value}-{current player credits} checkback is ignored for "sell options"(, as the price value is added).
- {Weight value}-{current free cargo space} checkback is ignored for "sell options"(, as that "sale standin" item should be removed directly after it has been awarded it to the ship. (At this point it is the responsibility of the OXP developers to remove it by script at once. This might be a source for bugs, though it worked fine in tests.)
- A given weight value in "sell options" is being used to create a "cargo space will be freed by X tons" message at the end of the description.

The current way for OXP developers to create a selling entry in the purchase menu, is to create standin items, holding the description text and linking to a script that will remove itself and the "sold" items and maybe even refunding credits. This method has proven to be powerful enough to lead to satisfying results, but it has shortcomings.
- Sale entries don't stand out as much as the fine repair option. That's a shame. :) 
- The sale entry's price will always be subtracted first, what makes a satisfiable presentation of refunded credits difficult.

The changes I propose enables developers to create distinguishable "Sale" entries and still using the old method.

### DIFF
--- a/Resources/Config/descriptions.plist
+++ b/Resources/Config/descriptions.plist
@@ -1656,9 +1656,12 @@
 	"equip-cash-value"				= "Cash: [credits|dcr].";
 	"equip-no-equipment-available-for-purchase" = "No equipment available for purchase.";
 	"equip-repair-@"				= " Repair:%@";
+	"equip-sell-@"					= " Sell:%@";
+	"upgradeinfo-@-price-is-for-refunding" = "%@ (Price will be refunded for the existing system).";
 	"upgradeinfo-@-price-is-for-repairing" = "%@ (Price is for repairing the existing system).";
 	"@-will-replace-other-energy"	="%@ (The installed energy recharge unit will be removed and sold for scrap).";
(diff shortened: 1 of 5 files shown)


ocz1:
Avoiding the {Weight value}-{current free cargo space} checkback doesn't create any problems, as the "sell option" item is never added to the ship, even when selected and "bought".

oocube:
I believe it would be nice if a general "equipment that can be bought can also be sold" option existed. Whether that makes sense is likely better judged by someone more into Ooniverse than me.

Is this a 'legit' PR or could the same be achieved via OXPs? Should we pull it to another branch so we can play around and test it?

oocube:
I merged this PR but then isolated it onto the branch PR172.
Now we have our build system work out the usual after which we can test and decide what to do.

Issue #176, opened 2016-02-05
Cache problems with rescaled models


AnotherCommander:
This issue is opened to track the bugs discovered during the discussions in the forums, seen here:
http://www.aegidian.org/bb/viewtopic.php?p=247369#p247369
In summary, it appears that writing  model data in the cache takes into account different scale factors for the same model as far as mesh data is concerned, but  does not take the same thing into account as far as octree data is concerned. The result is that, the octree data that happens to be written in the cache first gets applied to all subsequent creations of the same model, regardless of its scale. This creates some quite difficult to reproduce under normal conditions issues and messes up collision detection amongst other things.

Quoting from the discussion describing the problem in more detail:
--- Begin quote
To replicate: Use spara's shipdata.plist entry and shift-start Oolite, then launch from Lave. Using the debug console, spawn one normal alloy and take note of its mass: 
:spawn [alloy] 
T.mass 

The T is just a console shortcut, referring to the last entity that was spawned, in our case, a normal alloy. Now, spawn a test alloy: 

:spawn [alloy_test] 
T.mass 

You will find that the mass reported is the same for both alloys, despite the fact that one is 100 times bigger than the other. But we are not over yet. Try this now. Shift-restart Oolite again and this time spawn the big one first, then the normal one: 

:spawn [alloy_test] 
T.mass 
:spawn [alloy] 
T.mass 

This time, a huge number for mass is reported for both cases. So, what is going on? 

Well, the problem is that Oolite is pulling information about the model mesh and its associated octree, which are used for volume and mass calculations, from the cache. The cache has only one mesh entry for alloy, referring to the model name "alloy.dat" (btw, I was using the old alloy model, just to be sure that nothing weird is caused by the new one). So, whatever mesh data was written to the cache first, is what was going to be read in subsequent calls for alloys. If you had spawned a normal alloy first, you would have a normal alloy mesh and octree, if you had spawned the big one first, the cache would have a huge mesh and octree calculated for it and that's what it was going to use, even if you spawned a normal alloy or an alloy with a different scale factor. That's why in some cases there were awesome explosions happening with alloys, because their mass was something like five times the mass of a Cobra III. This also explains the notable FPS drops, because if a huge mass alloy is destroyed, it will probably spawn other alloys, which, if huge enough, will spawn others, etc. etc. - universe full.
--- End quote

[...]

PR #177, opened 2016-02-06
Proposed fix for issue #176 (Cache problems with rescaled models).

AnotherCommander:
This is the proposed resolution for the cache issues related to rescaling models (#176). Given that the cache is a high-risk area, I'd rather give a chance for the proposal to be reviewed first, rather than merging staight away.

As per this PR, separate entries are now inserted in the cache for models of different scale factors, applying both to mesh and octree data. Before this patch, octree data was entered in the cache for only one scale factor and subsequent calls for creation at other scale factors would result in a model that had collision detection and volume calculation issues. This could have very well been the cause of some weird bugs reported rarely in the past, whereby a player ship would be apparently hit by "something" and get ejected at excessive velocities towards a random direction.

The octree data is now tagged as "modelname.dat-XX.YYY", where XX.YYY is the scale factor referring to that particular model. Mesh caches were already getting tagged as separate entries for different scale factors, so now we bring also the octree tagging to work in the same manner. This resolves the main, quite serious, issue.

This PR resolves also a secondary issue, which is the filling up of the cache file with irrelevant data about randomly scaled, internally generated wreckage models. This kind of random internal rescaling happens only for wreckages at this point. The solution here was to implement an alternative rescaling method in ShipEntity, where a boolean parameter is passed for decdiding whether to write the result of the generated model to cache or not. With this, only wreckage models of scale 1.000 are written in the cache, unless specifically requested otherwise by means of script, debug console commands etc.. There is no reason to write data about e.g. wreck1,dat-1.803, when this model is randomly rescaled and chances of it being requested again any time soon are extremely slim. It's probably best to leave room for other objects, with better hit chances and it is certainly cleaner. This fix applies to both octree and mesh data.

I would also note that, before going with this solution, I had also looked at rescaling octrees using the already provided methods, However, this did not work out very well. With an octree at scale 1:1 in the cache, requesting a rescaled octree to 64 x original for example, is not the same as generating an octree for a 64x rescaled model from scratch.

We know that the fix outlined in this PR works on Windows, as a test build has been circulated in the forums and the initial reports seem to indicate good functionality and no performance hits. If this could be tested across the three supported platforms, it would be great.

### DIFF
--- a/Resources/Config/logcontrol.plist
+++ b/Resources/Config/logcontrol.plist
@@ -184,6 +184,7 @@
 	mesh.load.cached						= inherit;
 	mesh.load.uncached						= inherit;
 	mesh.load.octree.size					= no;
+	mesh.load.octreeCached					= inherit;
 	
 	mesh.load.error							= $error;
 	mesh.load.error.badCacheData			= inherit;
--- a/src/Core/Entities/OOEntityWithDrawable.h
+++ b/src/Core/Entities/OOEntityWithDrawable.h
@@ -32,6 +32,7 @@ MA 02110-1301, USA.
 @protocol OOSubEntity
 
 - (void) rescaleBy:(GLfloat)factor;
+- (void) rescaleBy:(GLfloat)factor writeToCache:(BOOL)writeToCache;
 
 // Separate drawing path for subentities of ships.
 - (void) drawSubEntityImmediate:(bool)immediate translucent:(bool)translucent;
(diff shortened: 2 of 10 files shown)


AnotherCommander:
I would like to go ahead and merge this fix to master. If there are no objections and/or possible better alternatives to the proposed above, I would consider doing the merge in a couple of days'
 time more or less.

PR #186, opened 2016-04-16
System information facility for all systems on a route.

kanthoney:
Facility to allow the player to obtain information for all tthe systems along the planned route.  In the F7 system information screen, use of the left/right arrow keys will display information for the previous/next system along the route.  The home key displays information for the current home system, and the end key displays information for the target system.

In the chart, the system for which information is displayed in F7 is highlighted with a blue circle.  This can be moved using Alt + left arrow/right arrow/home/end.


AnotherCommander:
Tested it and it seems to be working well. I am OK with merging this request to master.

PR #189, opened 2016-05-05
Mac fixup

JensAyton:
Boom.


AnotherCommander:
Hi Jens!

The branch compiles OK on Win32 and a (very) brief test run seems to indicate business as usual. I will try on Win64 tomorrow and report back, but I guess the only effective tests can happen post-merge, when this hits the nightlies. Speaking of which, do we need to do any changes to the terrastorage configuration to support the new XCode, if/when we merge this?

kanthoney:
Also compiles and runs on Linux.  And on Windows 7 64 bit.

AnotherCommander:
I think we are OK to merge this in. One of the commits fixes a crash too, which can only be a good thing. Jens, feel free to go ahead and merge anytime or I can do it later tonight if there are no objections.

While we have you here, would it be possible to re-enable the server functionality that provides the Mac nightly builds so that we won't need to rely on the terrastorage alternative Mac page?

Also, a proper Mac retina fix from someone who can actually test the result would be really welcome.

kanthoney:
I think the Retina fix is still in there, but a bit of code from an earlier attempt crept into master which is what's being excised.

PR #195, opened 2016-05-29
Travis CI

JensAyton:
* Sets up automatic integration builds on Mac OS X using Travis CI. This builds the project and the various Mac-specific bundles for each push and each pull request, and also runs the unit tests – all 13 of them.
* Converts readme to Markdown so we can have a build status icon in it.
* Updates the readme to vaguely reflect present reality – at least as seen from a Mac.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,5 @@
+language: objective-c
+osx_image: xcode7.3
+
+xcode_project: Oolite.xcodeproj
+xcode_scheme: 'Oolite - Test Release'
--- a/DebugOXP/Debug.oxp/Config/debugConfig.plist
+++ b/DebugOXP/Debug.oxp/Config/debugConfig.plist
@@ -13,7 +13,7 @@
 		:bgColor general "0.8 0.8 0.9"
 	
 	Settings like this are saved in Oolite's preferences and always override
-	settings in jsConsoleConfig.plist. An override can be removed in
+	settings in debugConfig.plist. An override can be removed in
 	JavaScript like this:
 	
 		delete console.settings["font-face"];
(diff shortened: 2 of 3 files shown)


JensAyton:
It would be good if some non-Mac people look at this commit in particular: https://github.com/OoliteProject/oolite/pull/195/commits/06218ce3941e598374c2e965350b30f8fb34839e

AnotherCommander:
I think it looks good. Just to be sure, am I right that it does not affect the other ports in any way?

JensAyton:
Yep. It will however e-mail you if you break the Mac build (with a link to [cleaned-up logs](https://travis-ci.org/OoliteProject/oolite/builds/133712139) and so forth).

kanthoney:
Could we use Travis for the Linux/Windows nightlies and give Getafix's modem a rest? I had a play at creating a docker container to create builds, but didn't get very far, particularly with cross compiling. Are the current Windows builds built natively or cross compiled?

AnotherCommander:
The Windows builds are built natively using a very heavily modified development environment (compared to standard MinGW distributions), including modified builds of support libraries, specifically for Oolite. Because of this, I am doubtful that travis integration would  work with them. The reason Getafix is able to provide Windows nightlies is that he is using an exact copy of the build environment on terrastorage. So for Windows I think that there is not much to do. Still, it would be great if we could utilize travis for Linux builds, since (usually) being able to build on Linux implies being able to build on Windows and vice versa.

JensAyton:
Objective-C is explicitly not supported out-of-the-box in Travis, but you may be able to hack it. Feel free to experiment. :-)

PR #198, opened 2016-07-10
Fix compile problems with gcc version 5

kanthoney:
Current master fails to build on GCC 5, which doesn't allow format strings in e.g. `OOLog` to be straightforward strings with no parameters.

### DIFF
--- a/src/Core/Debug/OODebugMonitor.m
+++ b/src/Core/Debug/OODebugMonitor.m
@@ -508,7 +508,7 @@ typedef struct
 
 - (void) dumpMemoryStatistics
 {
-	OOLog(@"debug.memStats", @"Memory statistics:");
+	OOLog(@"debug.memStats", @"%@", @"Memory statistics:");
 	OOLogIndent();
 	
 	//	Get texture retain counts before the entity dumper starts messing with them.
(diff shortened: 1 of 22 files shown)


AnotherCommander:
Is this security related? Because it certainly looks less readable than before.

I note that many submodules point to wrong versions. Is it possible to adjust this PR so that the submodules are pointed to their correct current revisions before we merge?

kanthoney:
I guess it's security related.

Actually, while my laptop (Debian with gcc-5.3.1-8) produces errors, my desktop (gcc-5.3.1-14) compiles fine, so it might be just that one compiler release.  I'm upgrading my laptop gcc version and will see if that fixes the problem.

JensAyton:
This looks like a straightforward compiler bug. It would be a security thing if the strings weren’t constants. What is the precise error?

kanthoney:
The error is 

    src/Core/Debug/OODebugMonitor.m:511:2: error: format not a string literal and no format arguments [-Werror=format-security]
      OOLog(@"debug.memStats", @"Memory statistics:");

It looks like the GNUstep config.make file on my laptop has picked up the line

    OBJCFLAGS= -g -O2 -fstack-protector-strong -Wformat -Werror=format-security

and that last flag is causing the error.  The corresponding line on my desktop doesn't have the same flag, which is why that compile works on that machine.

Closing this request.

PR #214, opened 2016-11-28
Adding additional system info to load screen

phkb:
Thanks to the assistance of another_commander, this code adds the economy, government and techlevel to the save game file, and then displays the appropriate graphical codes on the load screen when then save game is selected.


AnotherCommander:
Nice. You can try changing the part where it prints the data to something like what you had in your original patch for extra graphics effects:
```
if (locationTL)
{
	locationGov = [NSString stringWithFormat:@"%c", [cdr oo_unsignedCharForKey:@"current_system_government"]];
	locationEco = [NSString stringWithFormat:@" %c", (7 - [cdr oo_unsignedCharForKey:@"current_system_economy"]) + 16];
}
```

phkb:
OK, I think I'm starting to get the hang of this. Hopefully the latest commits put the graphical codes back in.

PR #217, opened 2016-12-07
Exposing the 'canDockShip' to Javascript to confirm if a ship can phy…

phkb:
…sically dock at a station


AnotherCommander:
With the `fitsInDock` method declaration moved to StationEntity.h, the private interface declaration around line 55 of StationEntity.m should be removed.
```
@interface StationEntity (OOPrivate)

- (BOOL) fitsInDock:(ShipEntity *)ship; // this should be removed, the method is not private anymore.
```

phkb:
OK, declaration removed!

AnotherCommander:
The declaration removed was from StationEntity, not PlayerEntity. I am also not sure why PlayerEntityControls.m was committed as part of this request either. The diff indicates whitespace difference in one line only, so I guess it's not a problem, but in the future please commit only what is relevant and necessary for the PR.

PR #220, opened 2017-03-16
Support for shader + texture atmosphere drawables simultaneously

AnotherCommander:
With the recent implementation of shader atmospheres, it became evident that the cloud effect (an internally generated texture) had to go, as the implementation was giving us only two options: either shader or texture atmosphere. The design also predicted that, if the shader atmosphere could not be generated for whatever reason, then there was an automatic fall-back to the texture  generated one.

This PR enables both atmosphere types to co-exist and I believe the visual result is worth it. The implementation is relatively simple and relies on two atmosphere drawables being defined per planet. One is for a texture and the other is for a shader effect. The design tries to keep as close as possible to the earlier one. Now, a texture atmosphere is always generated, like in 1.84. The shader generated drawable is also attempted and, if for whatever reason this cannot happen (e.g. uncompilable shader or gfx set to less than Extra Detail), then the texture atmosphere alone is used.

The PR contains some code that decides the radius of each of the drawables that may require further clarification. The idea is for the two drawables to co-exist, but if they both have the same dimensions there is visible interference between the two at the planet's edges. So, what we do is this: If both drawables are successfully created, then the shader one is set to the (standard) collision_radius + ATMOSPHERE_DEPTH, being the outmost, "primary" one and the texture drawable is set to just cover the planet. If the shader drawable is not created, then the texture one is set to the standard atmosphere radius, so that it looks the same as in 1.84. I believe this is OK for cosmetic reasons and did not indicate any gameplay problems during testing at various distances from the planet.

And this is what it looks like, with an ambient level setting of 0.15:
disoimprovedatmo01
laveimprovedatmo02
raleorimprovedatmo01

### DIFF
--- a/src/Core/Entities/OOPlanetEntity.h
+++ b/src/Core/Entities/OOPlanetEntity.h
@@ -41,6 +41,7 @@ MA 02110-1301, USA.
 @private
 	OOPlanetDrawable		*_planetDrawable;
 	OOPlanetDrawable		*_atmosphereDrawable;
+	OOPlanetDrawable		*_atmosphereShaderDrawable;
 	
 	BOOL					_miniature;
 	OOColor					*_airColor;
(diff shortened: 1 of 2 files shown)


AnotherCommander:
If there are no comments or suggestions, I'd like to go ahead and merge this PR at some time during the weekend.

PR #221, opened 2017-03-20
Rescaling experiment2

kanthoney:
Merge of current master into rescalingExperiment2.  There was one minor merge conflict involving whitespace. I don't want to push it direct to the branch in case that makes it awkward for Redspear.


AnotherCommander:
I had a few issues applying the PR diff,(some dependency control files were reported as not proper ones and I had to manually skip them) but that's almost certainly a problem from my part, since the PR is green for merge. It seems to work. I can't do the extensive testing Redspear does, but I think that we could go ahead with the merge, if he is OK with it too.

AnotherCommander:
Given the latest updates by Redspear on the forum, I am not sure that this PR is relevant anymore. Still, I think it should be merged just so that the branch becomes up to date. Redspear should really consider doing a fork and working on that, followed by a PR from his side when all is ready.

kanthoney:
I'm not at my computer at the moment, but I'll add in the changes when I get back.

kanthoney:
You're right, we don't need this any more, so I'm closing it.

PR #223, opened 2017-05-08
Perlin3D Planet and Atmosphere Textures as planetinfo.plist Option

AnotherCommander:
Rather than setting Perlin3D planet and atmosphere textures perference at compile time, I thought that it could be a better idea if we could have this selection as an option. This could motivate also contributors to try to optimize the Perlin3D code, too.

This pull request implements this idea. The type of texture for planets and atmospheres can be set via the "perlin_3d" key in planetinfo.plist for each planet separately or for everything, if set inside the "universal" block. 

By default we keep Perlin3D disabled due to its performance impact. There is an approximately two seconds delay on my system when viewing planets in the F7 screen with this enabled. Hyperspace transit takes somewhat longer too, as the new system textures get created. However, more powerful systems could maybe use this without getting too penalized, so it would be interesting to check the feature on more recent machines.


AnotherCommander:
There seems to be a little issue with the Perlin3D textures. It seems that they are not exactly procedural. Below are two images of the P3D textures generated for Celearen (G5) from two sequential runs of the game. You can re-create them for own testing purposes by setting DEBUG_DUMP in OOPlanetTextureGenerator.m to 1. As you see, they are not the same. Standard textures are procedural and appear the same no matter how many times the game is run.
image
image
This has been the case since the introduction of the P3D code and would need further investigation before P3D textures can be considered for default (some performance improvements would be very welcome too).

AnotherCommander:
If there are no comments or suggestions, I would like to go ahead and merge this PR during the weekend.

PR #224, opened 2017-05-23
F7 screen config and layout

phkb:
Opens the layout of the information on the F7 screen so that OXP's can potentially add their own information in the top part of the screen.

Also, I've created an OXP to add a control layer to the layout, which can be found here: [SystemDataConfig_0.3.zip](https://app.box.com/s/zl24eisa5orvrrsskzbh5q137rijs887)

### DIFF
--- a/Resources/Config/descriptions.plist
+++ b/Resources/Config/descriptions.plist
@@ -1538,17 +1538,35 @@
 	"sysdata-tl"					= "Tech Level:";
 	"sysdata-tl-value"				= "[techLevel]";
 	"sysdata-pop"					= "Population:";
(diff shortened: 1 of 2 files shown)


AnotherCommander:
I think it looks and works OK. Let's give it a while in case somebody would like to test further or comment, then I believe we can merge.

AnotherCommander:
OK, it's going in. @phkb: Feel free to put a few lines describing the feature in the forum Progress thread, if you want.

phkb:
Thanks for that a_c. I've just posted an update on "Progress".

PR #227, opened 2017-05-30
Adding 'addCargoEntity' method to player ship

phkb:
This adds the "addCargoEntity" to the player ship, allowing cargo entities to be added directly to the players cargo hold. 1 required param (the actual cargo entity, which must be a valid cargo entity currently in space somewhere), 2 optional params: processEvents (bool, default false) to indicate whether normal scooping events (eg shipScoopedOther) should be executed; processMessages (bool, default false) to indicate whether normal scooping messages (eg "You rescued Bill Gates")  should be displayed.

### DIFF
--- a/src/Core/Scripting/OOJSShip.m
+++ b/src/Core/Scripting/OOJSShip.m
@@ -2202,8 +2202,8 @@ static JSBool ShipAddCargoEntity(JSContext *context, uintN argc, jsval *vp)
 	OOJS_NATIVE_ENTER(context)
 	ShipEntity				*thisEnt = nil;
 	ShipEntity              *target = nil;
-	JSBool					proc_events = NO;
-	JSBool					proc_messages = NO;
+	JSBool					procEvents = NO;
+	JSBool					procMessages = NO;
 
 	GET_THIS_SHIP(thisEnt);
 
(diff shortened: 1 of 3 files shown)


AnotherCommander:
Not too happy with the crash fix, need to do some investigation once time permits. Please hold this PR open for now.

phkb:
Yes, it was a rather hasty fix during my homebound commute. It works (at least from the tests I've performed) but I knew I was missing something.

src/Core/Scripting/OOJSShip.m
```diff
@@ -2216,7 +2216,7 @@ static JSBool ShipAddCargoEntity(JSContext *context, uintN argc, jsval *vp)
 				   !JSVAL_IS_OBJECT(OOJS_ARGV[0]) ||
 				   !JSShipGetShipEntity(context, JSVAL_TO_OBJECT(OOJS_ARGV[0]), &target)))
 	{
```
AnotherCommander: A check against NULL JS value is needed here, otherwise something as simple as player.ship.addCargoEntity(player.ship.target, true, true) will throw a native exception.

src/Core/Scripting/OOJSShip.m
```diff
@@ -2231,17 +2231,28 @@ static JSBool ShipAddCargoEntity(JSContext *context, uintN argc, jsval *vp)
...
+		[thisEnt scoopUpProcess:target processEvents:YES processMessages:YES];
+	}
+	if (proc_events && !proc_messages) 
+	{
+		[thisEnt scoopUpProcess:target processEvents:YES processMessages:NO];
+	}
+	if (!proc_events && !proc_messages) 
+	{
+		[thisEnt scoopUpProcess:target processEvents:NO processMessages:NO];
+	}
```
AnotherCommander: All this added block must go away. Try instead:
```
OOJS_BEGIN_FULL_NATIVE(context)
[thisEnt scoopUpProcess:target processEvents:procEvents processMessages:procMessages];
OOJS_END_FULL_NATIVE
```

AnotherCommander:
OK, did a quick pass and added a few comments. There are a couple of things that need fixing.

PR #230, opened 2017-06-07
Fix Travis builds

JensAyton:
I _think_ this will fix it.


AnotherCommander:
The PR has passed its checks, it refers to OSX only and is very easy to revert in case of any problems. So, let's go ahead and merge it.

PR #232, opened 2017-06-12
Adds events and controls for MFDs and primed equipment

phkb:
This update hopefully provides the events and methods that will allow OXP devs to automatically prime a piece of equipment whenever an MFD is selected or changed by the player. There are two new events (selectedMFDChanged and mfdKeyChanged) and one new method (player.ship.setPrimedEquipment).

### DIFF
--- a/src/Core/Entities/PlayerEntity.m
+++ b/src/Core/Entities/PlayerEntity.m
@@ -7913,8 +7913,21 @@ NSComparisonResult marketSorterByMassUnit(id a, id b, void *market);
 
 - (BOOL) setPrimedEquipment:(NSString *)eqKey
 {
+	NSUInteger c = [eqScripts count];
+	NSUInteger current = primedEquipment;
 	primedEquipment = [self eqScriptIndexForKey:eqKey];	// if key not found primedEquipment is set to primed-none
-	return (primedEquipment != [eqScripts count]);
+	BOOL result = YES;
+	if (primedEquipment == c)
+	{
+		primedEquipment = current;
+		result = NO;
+	}
+	else 
+	{
+		NSString *equipmentName = [[OOEquipmentType equipmentTypeWithIdentifier:[[eqScripts oo_arrayAtIndex:primedEquipment] oo_stringAtIndex:0]] name];
+		[UNIVERSE addMessage:OOExpandKey(@"equipment-primed", equipmentName) forCount:2.0];
+	}
+	return result;
 }
 
 
--- a/src/Core/Entities/PlayerEntity.m
+++ b/src/Core/Entities/PlayerEntity.m
@@ -5376,6 +5376,7 @@ NSComparisonResult marketSorterByMassUnit(id a, id b, void *market);
 - (void) cycleMultiFunctionDisplay:(NSUInteger) index
 {
 	NSArray *keys = [multiFunctionDisplayText allKeys];
+	NSString *key = nil;
 	if ([keys count] == 0)
 	{
 		[self setMultiFunctionDisplay:index toKey:nil];
@@ -5384,20 +5385,27 @@ NSComparisonResult marketSorterByMassUnit(id a, id b, void *market);
 	id current = [multiFunctionDisplaySettings objectAtIndex:index];
 	if (current == [NSNull null])
 	{
-		[self setMultiFunctionDisplay:index toKey:[keys objectAtIndex:0]];
+		key = [keys objectAtIndex:0];
+		[self setMultiFunctionDisplay:index toKey:key];
 	}
 	else
 	{
 		NSUInteger cIndex = [keys indexOfObject:current];
 		if (cIndex == NSNotFound || cIndex + 1 >= [keys count])
 		{
+			key = nil;
 			[self setMultiFunctionDisplay:index toKey:nil];
 		}
 		else 
 		{
-			[self setMultiFunctionDisplay:index toKey:[keys objectAtIndex:(cIndex+1)]];
+			key = [keys objectAtIndex:(cIndex+1)];
+			[self setMultiFunctionDisplay:index toKey:key];
 		}
 	}
+	JSContext *context = OOJSAcquireContext();
+	jsval keyVal = OOJSValueFromNativeObject(context,key);
+	ShipScriptEvent(context, self, "mfdKeyChanged", INT_TO_JSVAL(activeMFD), keyVal);
+	OOJSRelinquishContext(context);
 }
 
 
@@ -5406,6 +5414,9 @@ NSComparisonResult marketSorterByMassUnit(id a, id b, void *market);
 	activeMFD = (activeMFD + 1) % [[self hud] mfdCount];
 	NSUInteger mfdID = activeMFD + 1;
 	[UNIVERSE addMessage:OOExpandKey(@"mfd-N-selected", mfdID) forCount:3.0 ];
+	JSContext *context = OOJSAcquireContext();
+	ShipScriptEvent(context, self, "selectedMFDChanged", INT_TO_JSVAL(activeMFD));
+	OOJSRelinquishContext(context);
 }
 
 
@@ -7897,6 +7908,12 @@ NSComparisonResult marketSorterByMassUnit(id a, id b, void *market);
 	{
 		return [[OOEquipmentType equipmentTypeWithIdentifier:[[eqScripts oo_arrayAtIndex:idx] oo_stringAtIndex:0]] name];
 	}
+}
+
+
+- (void) setPrimedEquipment:(NSString *)eqKey
+{
+	primedEquipment = [self eqScriptIndexForKey:eqKey];	// if key not found primedEquipment is set to primed-none
 }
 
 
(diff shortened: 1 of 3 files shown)


AnotherCommander:
Should player.ship.setPrimedEquipment be completely silent? I fear that scripts changing the primed equipment without any sort of notification might interfere with the gameplay experience. Especially if the equipment fails to prime for whatever reason and gets set to null, it might confuse a player who had already a different piece of equipment primed and now find that nothing is primed anymore. Maybe an optional second boolean parameter might be an idea?.

phkb:
You're right, that would be handy to know. New check-in covers this I think.

AnotherCommander:
Some observations from testing:
- The mfdKeyChanged event fires every time I toggle an MFD. Is this behaviour desired? Should it not fire only when we jump from MFD#0 to MFD#1 or vice versa?
- I am still not very happy with the setPrimedEquipment method in terms of gameplay. Imagine this scenario: I have the cloak primed and using it while in combat. Then a script decides to set a different primed equipment and fails for whatever reason. Suddenly and without any warning, my primed equipment status is now set to none primed and I find myself wondering why the cloak has stopped working mid-combat.

In general, we should be very careful whenever we empower scripts to act on things that are under direct player control. For the case of the setPrimedEquipment, I would consider two options and if anyone has more ideas, feel free to bring them up.
- Do not let the primed equipment go to none. Store in a local variable the current primed equipment and if setting it fails, then restore whatever equipment we originally had primed.
- Bring up a message (screen, log or both) to notify the player that something is changing.

phkb:
mfdKeyChanged was intended to fire every time an MFD is toggled. The purpose was to provide a way for an MFD that has a corresponding piece of primable equipment to know when it has been selected (either through the player cycling an MFD, or by the player selecting an MFD position) so it has the opportunity to automatically prime it's equipment item. 

Adding a visual notification of any change via setPrimedEquipment is a good idea, as is a better fall-back method.

Your point about setPrimedEquipment usage in gameplay is taken, though, and I'm not sure I have any good answer for it, either (other than having a large warning sign on the Wiki doco). I can only think of rather arbitrary reasons for limiting its use (eg only while docked, or condition green, or something like that), none of which make any sense or make it a valuable proposition. If you feel it's too dangerous then it's probably not worth doing.

I'll add the extra bits noted above, but we might shelve this idea for now.

src/Core/Entities/PlayerEntity.m
```diff
@@ -7913,8 +7913,21 @@ - (NSString *) primedEquipmentName:(NSInteger)offset
 
 - (BOOL) setPrimedEquipment:(NSString *)eqKey
 {
+	NSUInteger c = [eqScripts count];
+	NSUInteger current = primedEquipment;
 	primedEquipment = [self eqScriptIndexForKey:eqKey];	// if key not found primedEquipment is set to primed-none
-	return (primedEquipment != [eqScripts count]);
+	BOOL result = YES;
+	if (primedEquipment == c)
```
AnotherCommander: If we changed this to `if (primedEquipment == c && ![eqKey isEqualToString:@""])`, we could make the setPrimedEquipment method take care of un-priming everything too. See below proposed change to the message display line.

src/Core/Entities/PlayerEntity.m
```diff
@@ -7913,8 +7913,21 @@ - (NSString *) primedEquipmentName:(NSInteger)offset
...
+	{
+		primedEquipment = current;
+		result = NO;
+	}
+	else 
+	{
+		NSString *equipmentName = [[OOEquipmentType equipmentTypeWithIdentifier:[[eqScripts oo_arrayAtIndex:primedEquipment] oo_stringAtIndex:0]] name];
+		[UNIVERSE addMessage:OOExpandKey(@"equipment-primed", equipmentName) forCount:2.0];
```
AnotherCommander: As per above, this could be changed as follows to accomodate un-priming upon reception of empty string:
`[UNIVERSE addMessage:[eqKey isEqualToString:@""] ? OOExpandKey(@"equipment-primed-none") : OOExpandKey(@"equipment-primed", equipmentName) forCount:2.0];`

AnotherCommander:
I think it's pretty OK now. I've proposed a couple of small changes to allow the setPrimedEquipment method to unprime as well (it would do that if an empty string is sent instead of an equipment key). I think it's OK to merge, but the only other suggestion I'd have to make at this point is, how about a second parameter in the setPrimedEquipment method, to control whether the prime(/unprime) messages are printed? This would give complete control to the script and cover edge cases where the scripter would specifically want that the notifications be suprressed.

phkb:
OK, I think I've covered everything now. Let me know if I've messed up somewhere, though.

src/Core/Entities/PlayerEntity.m
```diff
@@ -7911,21 +7911,24 @@ - (NSString *) primedEquipmentName:(NSInteger)offset
...
 
-- (BOOL) setPrimedEquipment:(NSString *)eqKey
+- (BOOL) setPrimedEquipment:(NSString *)eqKey showMessage:(BOOL)showMsg
 {
 	NSUInteger c = [eqScripts count];
 	NSUInteger current = primedEquipment;
 	primedEquipment = [self eqScriptIndexForKey:eqKey];	// if key not found primedEquipment is set to primed-none
 	BOOL result = YES;
-	if (primedEquipment == c)
+	if (primedEquipment == c && ![eqKey isEqualToString:@""])
```
AnotherCommander: Suggest declaring a BOOL variable like unprimeRequested or any name you want, e.g. `BOOL unprimeRequested = [eqKey isEqualToString:@""];`, since you use the comparison to null string twice in the method.

src/Core/Entities/PlayerEntity.m
```diff
@@ -7911,21 +7911,24 @@ - (NSString *) primedEquipmentName:(NSInteger)offset
...
 	{
 		primedEquipment = current;
 		result = NO;
 	}
 	else 
 	{
-		NSString *equipmentName = [[OOEquipmentType equipmentTypeWithIdentifier:[[eqScripts oo_arrayAtIndex:primedEquipment] oo_stringAtIndex:0]] name];
-		[UNIVERSE addMessage:OOExpandKey(@"equipment-primed", equipmentName) forCount:2.0];
+		if (showMsg == YES)
```
AnotherCommander: Suggest modification to `if (primedEquipment != current && showMsg)` to avoid occurence of messages about priming equipment that is already primed at that time.

src/Core/Entities/PlayerEntity.m
```diff
@@ -7911,21 +7911,24 @@ - (NSString *) primedEquipmentName:(NSInteger)offset
...
+		{
+			NSString *equipmentName = [[OOEquipmentType equipmentTypeWithIdentifier:[[eqScripts oo_arrayAtIndex:primedEquipment] oo_stringAtIndex:0]] name];
+			[UNIVERSE addMessage:[eqKey isEqualToString:@""] ? OOExpandKey(@"equipment-primed-none") : OOExpandKey(@"equipment-primed", equipmentName) forCount:2.0];
```
AnotherCommander: Here is the second comparison of eqKey to empty string. The boolean proposed earlier can be used here.

src/Core/Scripting/OOJSPlayerShip.m
```diff
@@ -1503,13 +1503,15 @@ static JSBool PlayerShipSetMultiFunctionText(JSContext *context, uintN argc, jsv
...
-// setPrimedEquipment(key)
+// setPrimedEquipment(key, [noMessage])
 static JSBool PlayerShipSetPrimedEquipment(JSContext *context, uintN argc, jsval *vp)
 {
 	OOJS_NATIVE_ENTER(context)
 
 	NSString				*key = nil;
 	PlayerEntity			*player = OOPlayerForScripting();
+	JSBool					showMsg = YES;
+	BOOL					result = NO;
```
AnotherCommander: I think we can do without the result boolean here.

src/Core/Scripting/OOJSPlayerShip.m
```diff
@@ -1520,8 +1522,14 @@ static JSBool PlayerShipSetPrimedEquipment(JSContext *context, uintN argc, jsval
...
-	
-	OOJS_RETURN_BOOL([player setPrimedEquipment:key]);
+	if (argc > 1 && EXPECT_NOT(!JS_ValueToBoolean(context, OOJS_ARGV[1], &showMsg)))
+ 	{
+ 		OOJSReportBadArguments(context, @"PlayerShip", @"setPrimedEquipment", MIN(argc, 2U), OOJS_ARGV, nil, @"boolean");
+ 		return NO;
+ 	}
+
+	result = [player setPrimedEquipment:key showMessage:showMsg];
+	OOJS_RETURN_BOOL(result);
```
AnotherCommander: Can just use `OOJS_RETURN_BOOL([player setPrimedEquipment:key showMessage:showMsg]);`.

AnotherCommander:
I think it's pretty much good to go. I just made a few minor proposals. The one behaviour changing proposal is the one about avoiding priming messages for equipment that is already primed at the time of the method's execution.

phkb:
Thanks for the feedback, a_c. All improvements in now!

Issue #222, opened 2017-04-04
Oolite is still on SDL 1.2


alugarius:
I want SDL2!
I play Pioneer instead of Oolite because of this.

sjnewbury:
I'm thinking I'll take look into porting to SDL2.

In https://github.com/OoliteProject/oolite/issues/91 it was mentioned there are local changes to the in-tree binary libSDL-1.2, the Gentoo ebuild produces an Oolite binary linked to the system libSDL-1.2.so, maybe that explains the fullscreen vblank sync and performance issues with my build... nope it seems not.  Is there any more information on what changes have been made to the in-tree version?

AnotherCommander:
@sjnewbury Yes. All modifications to the standard libraries are listed here:
https://github.com/OoliteProject/oolite/blob/master/Doc/ExternalLibrariesSourceCodeChanges.txt

sjnewbury:
I've made a start.  Initially I'm just doing the bare minimum to get it working with the same functionality.  Later new features and capabilities are possible such as multiple screens OpenGL3/4 core profiles etc.

I don't know if the Windows changes noted above are still necessary with SDL2.  I'll create a pull request once I've tested it here on Linux, hopefully somebody with a Windows build setup can test it.

sjnewbury:
Some thoughts while working on this:

Splash screen could/(should?) be a separate window - I'm leaving the code as is at this stage for minimal invasiveness.

A lot of special casing for Windows is probably not necessary with SDL2.  I've removed some, but not all of this code.  It will need testing.  Probably most of it can be dropped.

There is a lot of code handling screen modes, I'm dropping it all on the floor.  The API has improved for SDL2 here and I don't think *any* of it is needed.  Unless I'm missing something...

"Surfaces" are a "compatible" feature with SDL2, to use GL you just need to create a GL context, so I've replaced "surface" with "glContext" thoughout, with the exception of the screenshot handling code, where I'm specifically creating a surface from the window for that function.  Hopefully, that will let me get away with leaving all that code alone.

HiDPI support should work on all supported platforms now, but I can't test it.  I'm trying not to break it!

sjnewbury:
I left in the screen mode support largely as is for now.  As I mentioned, I just want to alter as little as possible.

I've got the game running with "-nosplash".  Obviously buggy, since without "-nosplash" the window fails to appear on screen although it shows up in the window list.

[...]

PR #234, opened 2017-06-29
[RFC,WIP] Initial experimental SDL2 support

sjnewbury:
This is an initial (mostly working) port to SDL2.

By default SDL1.2 is still used, but by chaning the config.make
option, SDL2 is used instead as a build-time option.

For current status see commit message and issue #222

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -1,9 +1,21 @@
 include $(GNUSTEP_MAKEFILES)/common.make
 include config.make
 
-vpath %.m src/SDL:src/Core:src/Core/Entities:src/Core/Materials:src/Core/Scripting:src/Core/OXPVerifier:src/Core/Debug
-vpath %.h src/SDL:src/Core:src/Core/Entities:src/Core/Materials:src/Core/Scripting:src/Core/OXPVerifier:src/Core/Debug:src/Core/MiniZip
-vpath %.c src/SDL:src/Core:src/BSDCompat:src/Core/Debug:src/Core/MiniZip
+ifeq ($(OO_SDL2_ENABLED),yes)
+	OO_SDL_DIR               = src/SDL2
+        SDL_CFLAGS               = `sdl2-config --cflags` -DOO_ENABLE_SDL2
+        SDL_OBJCFLAGS            = `sdl2-config --cflags` -DOO_ENABLE_SDL2
+        SDL_OBJC_LIBS     = -lSDL2main -lSDL2
+else
+	OO_SDL_DIR               = src/SDL
+        SDL_CFLAGS               = `sdl-config --cflags`
+        SDL_OBJCFLAGS            = `sdl-config --cflags`
+        SDL_OBJC_LIBS     = -lSDLmain -lSDL
+endif
+
+vpath %.m $(OO_SDL_DIR):src/Core:src/Core/Entities:src/Core/Materials:src/Core/Scripting:src/Core/OXPVerifier:src/Core/Debug
+vpath %.h $(OO_SDL_DIR):src/Core:src/Core/Entities:src/Core/Materials:src/Core/Scripting:src/Core/OXPVerifier:src/Core/Debug:src/Core/MiniZip
+vpath %.c $(OO_SDL_DIR):src/Core:src/BSDCompat:src/Core/Debug:src/Core/MiniZip
 GNUSTEP_INSTALLATION_DIR         = $(GNUSTEP_USER_ROOT)
 ifeq ($(GNUSTEP_HOST_OS),mingw32)
     GNUSTEP_OBJ_DIR_NAME         := $(GNUSTEP_OBJ_DIR_NAME).win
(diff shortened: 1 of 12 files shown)


AnotherCommander:
I have not been able to install SDL2 on my dev computer yet, but just from eyeballing the changes in the PR, I can tell you that the Windows build is currently broken. There are still many references to SDL_Window scattered around in MyOpenGLView.m, which are certainly going to trigger a compile error since the SDL_Window definition is gone. All those references will have to be changed to Main_Window.

sjnewbury:
Strange, I thought I'd done a search and replace.  Will fix that!

sjnewbury:
@AnotherCommander Fixed the spurious references to SDL_Window.  Honestly, I am sure I had done that already at some point!

sjnewbury:
@AnotherCommander As far as I know the only significant outstanding issues/regressions with my Linux X11/Wayland tested build is the strange splash screen bug and fullscreen mode switching from smaller sizes upwards, while switching to windowed and back works.

sjnewbury:
Would be interesting to know if it works on Android.

KonstantinosSykas:
Hi. I'm maintaining the environment that produces the Linux distro independent Oolite installation packages.

The environment is an Ubuntu 8.04 (2.6.24-28 kernel) with a gcc v4.2.4. Pretty archaic, I know, but it has demonstrated to produce robust binaries with astonishing compatibility in the big ocean of Linux distros, making happy a lot of our followers having an old laptop with an old Linux distro, still managing to play Oolite decently!

Compiling libSDL2 on this environment breaks, complaining about missing mutlitouch device definitions.
Furthermore, warns about (among other things)
>pragma GCC diagnostic push/pop (old gcc)

and

>dereferencing type-punned pointer will break strict-aliasing rules

the second one leading to undefined behavior, from a quick search that I did.

I am not familiar with Gentoo, but do you get SDL2 source code that gets compiled in this enviroment or a latest binary from Gentoo repo? If it is source that you get, then can you share the ```configure``` and ```make``` logs?

sjnewbury:
@KonstantinosSykas Everything is compiled from source with Gentoo.  Giving you my logs wouldn't be much help though since I'm mostly bleeding-edge, my compiler is one of the few "stable" components gcc-6.3.0, so by your standards, that's almost from the future! ;-)

That's a *really* old kernel!

I'm not sure what versions are present with Ubuntu 8, but the Gentoo ebuild might well help you out with minimum versions and configure flags since Gentoo does try to keep supporting fairly "legacy" setups and will build with few dependencies if they aren't needed:
https://gitweb.gentoo.org/repo/gentoo.git/tree/media-libs/libsdl2/libsdl2-2.0.5.ebuild

sjnewbury:
@KonstantinosSykas Regarding the second GCC warning, I wouldn't be surprised if the lack of warning with 4.2.4 doesn't mean the behaviour is any different!

KonstantinosSykas:
My concern is that, at the end of the day, to be included in a release, the libSDL2 will have to come out from that environment, unless 
* the project team decides to set a baseline that will define the next gcc to be used, defining a backward compatibility level with older Linux distros and 
* someone will stand up and decide to setup the new development environment to support that decision.

Hmmm... food for thought for all of us. :wink:

sjnewbury:
@KonstantinosSykas Since it's a build time option, unless and until use is made of the SDL2 capabilities perhaps having two builds Would be the way to go? One for legacy systems and one for modern/alternative systems.  I'm tempted to set up an Android build environment to see if I can make a build that works on my phone.

sjnewbury:
Of course this would mean it wouldn't be feasible to implement features like multiple Oolite screen -> display mapping.

sjnewbury:
Since I changed from using the "native fullscreen" SDL_WINDOW_FULLSCREEN_DESKTOP to using the classic SDL_WINDOW_FULLSCREEN the size of the window gets changed on return to fullscreen.  With "native fullscreen" the old window size gets remembered.  I think it is necessary to remember the old dimensions and restore them on return from fullscreen, perhaps doing so also before switching fullscreen mode will help there too?  I think, looking at the windows code, it does something like that already.

sjnewbury:
Another data-point, under wayland, OpenGL ES is being used by default which has some interesting consequences with exhaust rendering (it's stippled) no doubt amongst other things.

AnotherCommander:
> I think, looking at the windows code, it does something like that already.

Yes, the way we do it on Windows is that we query the current desktop resolution and if the fullscreen mode we are going to has the same resolution, then we don't do a display mode change at all, we just modify the dimensions and the style of the game window to desktop size borderless. This has the benefit of being extremely smooth and fast, does not confuse any post-porcessing gfx injectors that might be in use and as a bonus, Windows automatically removes the taskbar without us having to do anything, as it sees an attempt to cover the entire screen with our window. If we detect that our target mode is of different size to our curent desktop, then we have to do a mode change (but the window is still set up as a borderless desktop-size so alt-tabbing etc. is still possible)

I still haven't been able to do anything with SDL2. Right now the priority is the message gui handling (told you, I prefer spending time on the game rather than its libraries :-) ).

alugarius:
Looks like I kicked of a something :)

laanwj:
> Another data-point, under wayland, OpenGL ES is being used by default which has some interesting consequences with exhaust rendering (it's stippled) no doubt amongst other things.

Heh, might want to add `SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_COMPATIBILITY);` or `SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);` before creating the window to request desktop GL.

oocube:
As of #620, which is merged to master already, we are on SDL3. This PR is no longer required.

PR #245, opened 2017-07-18
Adds handling of page up/down keys on most screens that can have mult…

phkb:
…iple pages


phkb:
The only screen I haven't touched yet is the "New game scenario" screen.

phkb:
OK, now all screens should be covered.

AnotherCommander:
Looks functional. Let's give it a day or two in case anyone discovers a bug that needs fixing or has a comment to add.

PR #247, opened 2017-07-21
Adds a "display_color" property to equipment items, used on equip ship screen

phkb:


AnotherCommander:
Seems OK. Any particular reason it's not read/write for scripts?

phkb:
Absolutely! The reasoning was, "Gosh I never thought of doing that!" :)
Seriously, I'd only really thought of this being used from within equipment.plist files, rather than within scripts. I'll have a good at adding this and see how it works.

AnotherCommander:
Will merge this one soon, if there are no objections or further comments. Tested it birefly, seems OK - and it leaves the damaged equipment color as it is, which is good.

PR #251, opened 2017-07-31
Adds get/set method to adjust values for screenbackgrounds.plist keys

phkb:
I don't know what I'm doing wrong, but I can seem to create a pull request from my fork that doesn't include all the pull requests I've done to rebase it. On the positive side, there are only three changed files now, which should mean that only the change I've requested should go in.

I'm actually wondering if I should delete my fork and recreate it, then do the pull request.


src/Core/Scripting/OOJSGlobal.m
```diff
@@ -424,6 +428,61 @@ static JSBool GlobalSetScreenBackground(JSContext *context, uintN argc, jsval *v
...
+		OOJSReportBadArguments(context, nil, @"getScreenBackgroundDefault", 0, OOJS_ARGV, nil, @"missing arguments");
+		return NO;
+	}
+	NSString		*key = OOStringFromJSValue(context, OOJS_ARGV[0]);
+	if (EXPECT_NOT(key == nil || [key isEqualToString:@""]))
+	{
+		OOJSReportBadArguments(context, nil, @"getScreenBackgroundDefault", 0, OOJS_ARGV, nil, @"key");
+		return NO;
+	}
+	id descriptor = [UNIVERSE screenTextureDescriptorForKey:key];
```
AnotherCommander: For clarity, this should be NSDictionary * rather than id.

PR #256, opened 2017-09-06
Adds primedEquipment property and playerChangedPrimedEquipment event

phkb:


AnotherCommander:
It works OK, but there is a design issue here. I think this is the only property that is read-only, but can be set via the player.ship.setPrimedEquipment(eqKey, showMsg) method. Normally one would expect any changeable property to be read/write, so we should be able to do something like player.ship.primedEquipment = "EQ_CLOAKING_DEVICE" and set it like this.

Of course, this leaves us with the question of what to do about the show message requirement. I was thinking about doing the property change without a message and use the setPrimedEquipment method when we want to be specific about the message shwoing or not, but this is likely complicating somewhat the JS interface for the player ship (read/write property + JS method for doing more or less the same thing). In an ideal world we should not need the method if the property is read/write. Any ideas about how we should go about this?

phkb:
Maybe what we do is make the "primedEquipment" property read/write, and default the write method to always show a message. But we also leave the "setPrimedEquipment" method as is, which then allows for the optional parameter to either show or hide messages. Alternatively we change this method name to be "primedEquipmentSilent", or something like that, and default it to not send messages.

phkb:
And to add to the confusion - have I broken travis? It's been a couple of days since I checked in the code, but it's still showing as being in progress.

phkb:
I've checked in my suggestion for comment.

phkb:
And Nick puts his foot in it again by suggesting exactly what a_c had suggested. D'oh! I've edited it away so my embarrassment won't be perpetuated into the future, but I'm sorry for not reading things clearly before I comment.

src/Core/Scripting/OOJSPlayerShip.m
```diff
@@ -724,6 +724,15 @@ static JSBool PlayerShipSetProperty(JSContext *context, JSObject *this, jsid pro
 			}
 			break;
 
+		case kPlayerShip_primedEquipment:
+			sValue = OOStringFromJSValue(context, *value);
+			if (sValue != nil)
+			{
+				[player setPrimedEquipment:sValue showMessage:YES];
+				return YES;
```
AnotherCommander: I think we should be able to return also NO here, in case of failure to set primed equipment. Since -setPrimedEquipment: returns BOOL anyway, I would simply go with
```
return [player setPrimedEquipment:sValue showMessage:YES];
```

AnotherCommander:
Just one observation in the latest commit proposal, as per above.

Regarding Travis, I have no idea what's holding it. I hope that things will be cleared once we merge with master, because I can't see anything wrong in the commits themselves.

AnotherCommander:
> Alternatively we change this method name to be "primedEquipmentSilent", or something like that, and default it to not send messages.

This sounds like a nice alternative. We could probably go with that, if no other suggestions come up.

phkb:
Should I make `primedEquipmentSilent` a method (ie just rename `setPrimedEquipment`), or make it a write-only property (if that's a thing in ObjC)? Just thinking about usage: if the default is `primedEquipment = "EQ_CLOAKING_DEVICE"`, it may make it more natural to do `primedEquipmentSilent = "EQ_CLOAKING_DEVICE"`, rather than `primedEquipmentSilent("EQ_CLOAKING_DEVICE")`

AnotherCommander:
What I think might be a possibility would be one of the two options:
- Keep only the read/write property with defaulting the priming message to NO. Then, leave it to the OXP creators to add the priming message themselves if they so wish.
- Include the priming message request in the equipment string itself using some predefined convention. For example, we could do `primedEquipment = "EQ_CLOAKING_DEVICE"` and prime with the message showing or do `primedEquipment = "EQ_CLOAKING_DEVICE::OO_SILENT"` to prime the equipment without showing the message. In this case what we'd have to do is check the equipment string for the ::OO_SILENT suffix and act accordingly.

If none of the above is considered good enough, I think I'd rather prefer the method way, since I can't think of any write-only properties so far. Not that it would be impossible (not really sure till we try it), but it's probably best if only already existing ways to do things are used (plus it's simpler just to rename the method to setPrimedEquipmentSilent).

cim--:
Write-only properties are probably not a good idea - the method way seems cleanest and most consistent with what's already there.

phkb:
I've changed the `primedEquipment` property to (a) return the result of the call to `setPrimedEquipment`, and (b) to not send a console message. Because of this of this last bit, there didn't seem to be any reason to change the name of the method to `setPrimedEquipmrentSilent`, as the property is now the silent one, and the default of the `setPrimedEquipment` is "YES", so there seems to be a natural delineation between the property and the method.

phkb:
And on the good news side, travis has apparently got over his fit of pique and is working again. Weird.

AnotherCommander:
If there are no further comments or ideas, I think we can go ahead and merge this in its current form.

PR #257, opened 2017-09-15
making playership.compassTarget readwrite

phkb:
Open to suggestions here. I believe this code works as intended, but there may be an easier way of doing this than what I've implemented here. I was trying to avoid having to re-write all the compass mode code into a new function, and this seemed like the simplest way to achieve that. Let me know what you think.

### DIFF
--- a/src/Core/Scripting/OOJSPlayerShip.m
+++ b/src/Core/Scripting/OOJSPlayerShip.m
@@ -699,15 +699,7 @@ static JSBool PlayerShipSetProperty(JSContext *context, JSObject *this, jsid pro
 					[player setNextCompassMode];
 					[player validateCompassTarget];
 				}
-				if ([player compassTarget] == current)
-				{ 
-					OOJSReportWarning(context, @"Compass target not found.");
-					return NO;
-				}
-				else
-				{
-					return YES;
-				}
+				return [player compassTarget] != current;
 			}
 			else 
 			{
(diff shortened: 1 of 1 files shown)


src/Core/Scripting/OOJSPlayerShip.m
```diff
@@ -678,7 +679,41 @@ static JSBool PlayerShipSetProperty(JSContext *context, JSObject *this, jsid pro
...
+					[player validateCompassTarget];
+				}
+				if ([player compassTarget] == current)
+				{ 
+					return NO;
+				}
+				else
+				{
+					return YES;
+				}
```
AnotherCommander: Lines 702 ti 706 can be simplified to 
`return [player compassTarget] != current;`

AnotherCommander:
Compass changes should be done very carefully. Not sure if I'll be able to test any time soon, so I trust that it has been tested and found working. Will leave it open for a little longer in case someone wants to do more testing or review, then I think we can merge.

AnotherCommander:
It seems I was in a bit of a hurry yesterday and failed to notice that there was a warning message printed in the part that was simplfied. Nick, if we can live without that message then the simplification is OK, but if we need it printed, then it's not and it should be OK to revert. Up to you.

phkb:
I think the simplified way is probably more consistent with other functions that either return true if it succeeds, and false if it doesn't, without having a specific log message to explain why. I'll add some text to the wiki documentation to explain the fail case.

AnotherCommander:
If noone has anything more to add, this is getting committed to master tomorrow.

PR #266, opened 2017-10-09
Implements MyOpenGLView::resetMouse for Mac

jobi-wan:
This recenters the virtual joystick and warps the mouse cursor to the center of the screen, in order to have the same functionality as its SDL equivalent resetMouse method.


AnotherCommander:
Excellent! Committing to master.

PR #267, opened 2017-10-12
Prevents recalculating route every frame on Chart view

jobi-wan:
This completes the code that was already in place to prevent unnecessary recalculation of routes.


AnotherCommander:
Judging by the variable names, it seems like the intention has always been for them to be static. It looks good, but I'd like to give it a test once I get a chance first.

jobi-wan:
Sure. The first commit introduced the bug where the route was not recalculated on nav mode change. Fixed by the 2nd. If there are other conditions than source, destination, mode change, under which the route should be recalculated, then that probably doesn't happen now.

PR #272, opened 2017-10-27
Implements mouse wheel for Mac

jobi-wan:
This implementation is pretty much identical to the SDL side.

http://www.aegidian.org/bb/viewtopic.php?f=6&t=18927

### DIFF
--- a/src/Core/GameController.m
+++ b/src/Core/GameController.m
@@ -315,6 +315,7 @@ static GameController *sSharedController = nil;
 
 - (void) performGameTick:(id)sender
 {
+	[gameView pollControls];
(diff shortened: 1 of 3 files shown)


src/Core/GameController.m
```diff
@@ -315,6 +315,7 @@ - (void) beginSplashScreen
 
 - (void) performGameTick:(id)sender
 {
+	[gameView pollControls];
```
AnotherCommander: I think this still needs an #if !OOLITE_SDL guard. We should not have to execute -pollControls: here in the SDL builds.

PR #273, opened 2017-10-27
Scripting: Makes pitch/roll/yaw for playerShip writable

jobi-wan:
In a discussion about the reverse-control behavior, it was suggested that pitch/roll/yaw for playership be writable. 
http://www.aegidian.org/bb/viewtopic.php?f=6&t=9142

### DIFF
--- a/src/Core/Entities/ShipEntity.m
+++ b/src/Core/Entities/ShipEntity.m
@@ -8626,54 +8626,60 @@ NSComparisonResult ComparePlanetsBySurfaceDistance(id i1, id i2, void* context)
 
 - (void) increase_flight_roll:(double) delta
 {
-	if (flightRoll < max_flight_roll)
-		flightRoll += delta;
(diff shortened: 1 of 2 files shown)


AnotherCommander:
Bug: large negative values for pitch, yaw and/or roll make everything go haywire. Example `player.ship.pitch=-500`

Error checking (or, alternatively, bounds checking) is needed.

jobi-wan:
The increase/decrease_flight_pitch/roll/yaw methods in ShipEntity.m assume that delta is always positive. The increase_ functions only check upper bounds; the decrease_ functions only check lower bounds. 

I can change the new code in OOJSPlayerShip.m to always pass positive deltas, 
or change those increase_/decrease_ methods in ShipEntity.m to cope well with negative deltas. 

I could reproduce the issue using your recipe. Both of these approaches fix it. I would suggest the latter.

PR #277, opened 2017-11-03
Bumps master branch version number to 1.87

jobi-wan:
Should master be labeled 1.87 now, so that OXPs that use new features can 'require' it?


jobi-wan:
None of them have changed encoding or line endings. 
I did a side-by-side compare on hexdumps of each. The only bytes that changed are sixes that became sevens. 
(By the way: InfoPlist.strings is the only one that is UTF-16)

AnotherCommander:
Thanks for the confirmation, merging.

PR #279, opened 2017-12-06
Windows Resources Addition & Manifest Embedding

AnotherCommander:
This PR is Windows-specific only, but I would like to have it tested before commiting, because it touches files used by the Linux build process, but also because it is a different way for handling some particular OS requirements of the Windows port of the game.

A set of resource files for Windows have been added to the build process. They should be usable and accessible by the Windows build only. Linux should be able to continue to build unaffected. These resource files at this time contain information about the icon to use for the executable and, more importantly, the application manifest, which Windows reads and acts on as required.

The embedded manifest tackles two previous issues at the same time and allows us also to simplify the core code quite a bit. The two issues are the high-DPI support and the reporting of the OS version we are running on. Both of them are already solved in master, but at the expense of loading and executing kernel functions at runtime to do the work for us. A fine solution, but maybe less elegant than we would like. The manifest is the Microsoft recommended way of doing this and it leaves the door open for potential future settings to be handled therein. Microsoft is using manifests quite a lot from what it seems and this may be the trend in the future, so supporting this kind of set up is probably a good idea overall.

And of course, we get a nice icon on our Oolite executable as a bonus (yay). 

So, please check at your own convenience and leisure and confirm that:
- The Linux build process is unaffected
- High-DPI support works just like before (all previous high-DPI code has been stripped in this PR and there are no reports like "SetProcessDpiAwareness(PROCESS_PER_MONITOR_DPI_AWARE) loaded and executed" - it needs to be tested with various DPI settings and on multiple monitors
- Windows 8 and higher versions are reported correctly in the log file (example: "Opening log for Oolite version 1.87 (x86-64 test release) under Windows 10.0.16299  64-bit at 2017-12-06 ...").


kanthoney:
I can confirm that it builds and runs under Linux.

phkb:
I can confirm a solid Win10 64bit build
`Opening log for Oolite version 1.87 (x86-64 test release) under Windows 10.0.15063  64-bit at 2017-12-07 08:27:03 +1100.`

AnotherCommander:
Thanks guys, that's good. I guess the multi-monitor DPI stuff will be a little bit trickier to test, let's give it some more time. In the meantime I have successfully run the game on Win 10 single momitor with 150% scaling; all OK thus far.

KonstantinosSykas:
Nightly environment for Linux32 and Linux64 also builds as usual. 
Didn't check execution; @kanthoney's check suffices! 😉

phkb:
Did another test with a 2nd monitor running 125% DPI, was able to start Oolite and move it onto the screen OK. Switched to make that screen my primary one, reran Oolite OK. 64bit in both cases. Any other specific 2nd monitor tests you like to see?

AnotherCommander:
@phkb The test you did is fine, thanks for that, I just wanted to ask if it would be possible to repeat it at 150% scaling. !50%  is where the OS DPI virtualization kicks in and if that works OK, it would be a guarnatee that we've got it correct.

KonstantinosSykas:
Tested as requested with multi monitor setup (2560x1440 & 1920x1080) at 175%. 
Behavior is same as the master branch. 
Seems good to GO Commander!

AnotherCommander:
Thanks all for your time and tests performed. Merged and closed.

PR #282, opened 2017-12-13
Multipage primable equip

phkb:
Adds multi-page support to the primable equipment manager.


AnotherCommander:
Very cool, almost there. The only issue I see is described below:
Here I am testing with 22 or so fictional items. This is the first screen I get, all good so far:
image

Next page shows the remaining items, but as you can see, Next Page is still enabled, while it should be greyed out.
image

I can select Next Page and it takes me to this screen:
image
This screen is not necessary (I could have selected Cloaking Device already from the previous one) and, what's more, Next Page is still available. I can click it say, ten times, and it will take me to the same screen ten times. If I start clicking Previous Page, it will take me back ten times before I see again the middle screen.

I think that once Next Page is set to grey out when the remaining items are enough to fill the UI screen, we can merge this.

AnotherCommander:
Nope. That doesn't work. The Next and Previous page selections are always selectable, even when greyed out. And if you select them when greyed out it's exception party everywhere. The problem is that the code does not understand "selectable", it does understand "unselectable" though. Here is the code that makes it work, with some simplifications, starting at line 202 of the script:
```
if (addNextPrev) 
{
	choice = {
		text: expandMissionText("oolite-primablemanager-next"), 
		unselectable: end >= equipment.length - 1, 
		color: end >= equipment.length - 1 ? "grayColor" : "yellowColor"
	};
	choices["ZZZZZZZ_OOLITE_1_NEXT"] = choice;
	choice = {
		text: expandMissionText("oolite-primablemanager-previous"), 
		unselectable: start <= 0, 
		color: start <= 0 ? "grayColor" : "yellowColor"
	};
	choices["ZZZZZZZ_OOLITE_2_PREV"] = choice;
}
```
I think we are good to go once this fix has been committed.

phkb:
I think the previous version of the fix was checked in by my evil twin brother. Or something.

PR #284, opened 2017-12-20
Adds extra chart configuration options for mission screens

phkb:
Adds "SHORT_RANGE_CHART_SHORTEST" and "SHORT_RANGE_CHART_QUICKEST" to the backgroundSpecial list. Also adds "CUSTOM_CHART", "CUSTOM_CHART_QUICKEST", and "CUSTOM_CHART_SHORTEST" backgroundSpecial types. These types make use of runScreen options "customChartZoom", "customChartCentre" and "customChartCentreInLY" to allow for a completely customised mission chart screen.


AnotherCommander:
This one looks good. @phkb , is there any minimal OXP by any chance, for being able  to test the new additions easily?

phkb:
I’ll put one together. Stand by. Watch this space.

phkb:
OK, here's an OXZ that will demonstrate the new features. Load any game, with or without the ANA. New F4 screeen (Unit Tests), with two options: "Short Range Chart" and "Custom Chart". Both will select a random system and show a map. The SRC will pick a system that has two paths, depending on the route mode. The CC will pick a system somewhere between 10 and 20 LY away. Same system will be shown while you remain in the screens. Exiting back to the F4 Interfaces screen will reset the systems so another system will be chosen when you go in. 
UnitTests.oxz: [https://app.box.com/s/82fn2i3j47c7dapg3fo8j74bvg0xx74r](url)

AnotherCommander:
There seems to be a rather minor bug with the OXZ (always selects the same planet), but I think it showcases the feature quite well. I think it is OK to go in, will probably pull it tomorrow if there are no objections.

PR #287, opened 2018-02-08
Rescaling Experiment IV

Redspear-of-Oolite:
Needs sun distance modifer setting (planetinfo.plist was too big for me to edit here).
I tested and recommend a value of  6.6
Think I've added everything... we'll see...


AnotherCommander:
Here is a link to the planetinfo.plist with the sun_distance key values already multiplied by 6.6: https://drive.google.com/open?id=1TrRpiC09mAvSnyGfpCwOW0NuMJDQ9a0f

Feel free to submit it to this PR.

Redspear-of-Oolite:
I was a bit messy in doing so but I think I've added the planetinfo changes. Thanks a_c.

mcarans:
@Redspear-of-Oolite Are you still around? If so, please can you tell me what this PR is about.

mcarans:
@oocube @phkb do you have any idea what this PR fixes or solves? I have asked the author but no response. If not, I suggest closing.

phkb:
The thing this OXP tried to "fix" is the sense of scale. It changes the sizes of planets, suns and ships to make things less obviously out of scale. However, it's a pretty big change, and even if we were to merge this, I'd want to do it in a way that would allow it to be turned on/off easily (either via JS and so OXPable, or via a setting in oolite.plist).

mcarans:
> The thing this OXP tried to "fix" is the sense of scale. It changes the sizes of planets, suns and ships to make things less obviously out of scale. However, it's a pretty big change, and even if we were to merge this, I'd want to do it in a way that would allow it to be turned on/off easily (either via JS and so OXPable, or via a setting in oolite.plist).

Currently the many values that need changing are hard coded and it looks to me like it would be quite difficult to make it configurable at runtime. I can see how it could be done at compile time using #ifdef or #if for Obj-C code and probably just duplicating the plist files with changed values swapping them into the build if desired but that would be little better than rebasing this PR. 

Do you know if anyone previously tested this PR and gave positive feedback about it? I'm reluctant to spend time on it given I wasn't involved in writing it or the previous rescaling experiments and if there isn't a strong reason to do it now given that there are other probably more important issues to address (eg. Vulkan, JS). However, this was one of the PRs blocking reformatting all the code consistently.

phkb:
A few people did tests. a_c created a Windows build based on it which he made available for download, but I don't know if that exists anymore. This is one of the things I'd like to see added to the game, but I think it can wait until after all the other changes we've got planned - it will need a stable core code to make the inevitable go-around on testing work successfully, I think.

mcarans:
> A few people did tests. a_c created a Windows build based on it which he made available for download, but I don't know if that exists anymore. This is one of the things I'd like to see added to the game, but I think it can wait until after all the other changes we've got planned - it will need a stable core code to make the inevitable go-around on testing work successfully, I think.

I have successfully (as far as I can tell) rebased this PR in my fork here: https://github.com/mcarans/oolite/tree/pr-287 since I don't have write access to the owner's fork. Packages are here: https://github.com/mcarans/oolite/releases/tag/1.93.1-PullRequest287.7992. Looks like planet is bigger on leaving space station and sun is smaller.

Please let me know if this does what you expect. What needs to be tested? Since I've managed to rebase this, maybe we can get it over the line this time.

oocube:
Maybe offer it on the forum. There are quite some threads containing "rescaling experiment".
We could also discuss the fact that such a change would actually change Oolite in a non-configurable way so people would have to decide which way to go.

mcarans:
> Maybe offer it on the forum. There are quite some threads containing "rescaling experiment". We could also discuss the fact that such a change would actually change Oolite in a non-configurable way so people would have to decide which way to go.

Good idea! I have posted there. I also PMed @Redspear-of-Oolite in the hope they'll take a look.

PR #288, opened 2018-02-09
Adds player.ship.chartHighlightMode property, and chartHighlightModeC…

phkb:
…hanged event


src/Core/Scripting/OOJSPlayerShip.m
```diff
@@ -644,6 +650,36 @@ static JSBool PlayerShipSetProperty(JSContext *context, JSObject *this, jsid pro
 			}
 			break;
 		
+		case kPlayerShip_chartHightlightMode:
```
AnotherCommander: I don't like this here. It would probably be much better to have this set up in a table in the Core/Tables folder and have OOConstToString.m handle both value to string and string to value conversions. This would allow us to add more modes in the future with only needing to update the table file, rather than adding code here.

What I would try to do is this:
- -Set up the table of LRC modes like below and remove the OOLRC_MODE enum from the headers
```
ENTRY(OOLRC_MODE_UNKNOWN,		-1)
ENTRY(OOLRC_MODE_SUNCOLOR,		0)
ENTRY(OOLRC_MODE_ECONOMY,		1)
ENTRY(OOLRC_MODE_GOVERNMENT,		2)
ENTRY(OOLRC_MODE_TECHLEVEL,		3)
```
- Change the OOStringFromLongRangeChartMode function as seen earlier.
- Add the reverse OOChartModeFromString in OOConstToString.m, defined like the other functions there that take a string and return an integer.
- Rewrite the case statement to maybe something like this:
```
case kPlayerShip_chartHightlightMode:
	sValue = OOStringFromJSValue(context, *value);
	if (sValue) 
	{
		OOLongRangeChartMode chartMode = OOChartModeFromString(sValue);
		if (chartMode > OOLRC_MODE_UNKNOWN)
		{
			[player setLongRangeChartMode:chartMode];
			[player doScriptEvent:OOJSID("chartHighlightModeChanged") withArgument:OOStringFromLongRangeChartMode([player longRangeChartMode])];
			return YES;
		}
		else
		{
			OOJSReportError(context, @"Unknown chart hightlight mode specified - must be either OOLRC_MODE_SUNCOLOR, OOLRC_MODE_ECONOMY, OOLRC_MODE_GOVERNMENT or OOLRC_MODE_TECHLEVEL.");
		}
	}
	return NO;	// not reachable if successfully set
	break;
```
Note that this would rename the CHART_MODE_* labels to OOLRC_MODE_*. Note also that we consider an unknown default mode for the chart, which should be reset to OOLRC_MODE_SUNCOLOR at startup. Note finally that in all the above I have changed OOLRC_MODE_NORMAL to OOLRC_MODE_SUNCOLOR, because I think it is more descriptive.

src/Core/OOConstToString.m
```diff
@@ -69,6 +69,17 @@
...
+{
+	switch (value)
+	{
+		case OOLRC_MODE_NORMAL: return @"CHART_MODE_SUNCOLOR";
+		case OOLRC_MODE_ECONOMY: return @"CHART_MODE_ECONOMY";
+		case OOLRC_MODE_GOVERNMENT: return @"CHART_MODE_GOVERNMENT";
+		case OOLRC_MODE_TECHLEVEL: return @"CHART_MODE_TECHLEVEL";
+	}
+	return @"UNDEFINED";
+}
```
AnotherCommander: I would rather change this to something like
```
NSString *OOStringFromLongRangeChartMode(OOLongRangeChartMode value)
{
	switch (value)
	{
		#include "OOLongRangeChartMode.tbl"
	}
	
	return @"UNDEFINED";
}
```
Remember that the ENTRY macro is defined at this point, so it is convenient to use it like the neighboring functions do. See below for a possible definition of the table.

phkb:
I think I covered everything. Adding new file... whoa, scary!

src/Core/Scripting/OOConstToJSString.h
```diff
@@ -143,6 +143,33 @@ OOINLINE OOCompassMode OOCompassModeFromJSValue(JSContext *context, jsval value)
 }
```
AnotherCommander: From here...

src/Core/Scripting/OOConstToJSString.h
```diff
@@ -143,6 +143,33 @@ OOINLINE OOCompassMode OOCompassModeFromJSValue(JSContext *context, jsval value)
...
+	return (OOLongRangeChartMode)OOConstantFromJSStringPRIVATE(context, string, &gOOLongRangeChartModeConstTable, kOOLongRangeChartModeDefault);
+}
+
+
+OOINLINE OOLongRangeChartMode OOLongRangeChartModeFromJSValue(JSContext *context, jsval value)
+{
+	extern struct ConstTable gOOLongRangeChartModeConstTable;
+	return (OOLongRangeChartMode)OOConstantFromJSValuePRIVATE(context, value, &gOOLongRangeChartModeConstTable, kOOLongRangeChartModeDefault);
+}
+
```
AnotherCommander: ... to here; Do we need these functions? You don't seem to be using them anywhere.

AnotherCommander:
Where are these two functions used?
- OOINLINE JSString *OOJSStringFromLongRangeChartMode(JSContext *context, OOLongRangeChartMode value)
- OOINLINE jsval OOJSValueFromLongRangeChartMode(JSContext *context, OOLongRangeChartMode value)

I think they can be removed too.

PR #294, opened 2018-05-02
Mousewheel handling

AnotherCommander:
I would like to incorporate the new mousewheel handling into master, but I need some help with testing on Linux and Mac. @OoliteProject/developers Please let me know if this implementation is deemed satisfactory. The request is _not_ expected to be merged, unless both Linux and Mac positive reports have been received.


KonstantinosSykas:
Linux 64bit and Mac basic testing seems fine. I think it's a GO!

Noticed that, in Mac, rolling the mouse-wheel towards the player increases the speed, while for Windows and Linux it decreases it. Nothing to do with your changes. The same behavior is in master branch too.

PR #302, opened 2018-07-02
Edit gui color settings in gui-settings.plist

phkb:
I haven't added functions for all the different value types in "gui-settings.plist" - just the color ones for the moment. And I'm not sure about my method of checking if the setting key is color-related. There might be a better way of doing this.

### DIFF
--- a/src/Core/GuiDisplayGen.m
+++ b/src/Core/GuiDisplayGen.m
@@ -432,6 +432,7 @@ static BOOL _refreshStarChart = NO;
 	}
 	[guiUserSettings release];
 	guiUserSettings = [guiCopy copy];
+	[guiCopy release];
 }
 
 
--- a/src/Core/GuiDisplayGen.m
+++ b/src/Core/GuiDisplayGen.m
@@ -419,6 +419,22 @@ static BOOL _refreshStarChart = NO;
 }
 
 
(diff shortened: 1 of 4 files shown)


src/Core/GuiDisplayGen.m
```diff
@@ -419,6 +419,22 @@ - (void) setGLColorFromSetting:(NSString *)setting defaultValue:(OOColor *)def a
...
+		[guiCopy removeObjectForKey:key];
+	} 
+	else
+	{
+		[guiCopy setObject:col forKey:key];
+	}
+	[guiUserSettings release];
+	guiUserSettings = [guiCopy copy];
+}
+
```
AnotherCommander: I would check this part again. I think there is a memory leak the way it is  written above.

src/Core/Universe.m
```diff
@@ -6305,6 +6305,7 @@ - (void) setScreenTextureDescriptorForKey:(NSString *)key descriptor:(NSDictiona
 	}
 	[screenBackgrounds release];
 	screenBackgrounds = [sbCopy copy];
+	[sbCopy release];
 }
```
AnotherCommander: This one should be OK but it is not related to this PR. It should be submitted separately (and will needconfirmation that it still works as expected and without crashes).

PR #311, opened 2018-10-15
fix issue where Station JS AI was not triggering an alertConditionCha…

phkb:
…nged event


phkb:
I believe this PR brings the JS AI in line with the previous plist AI, so that the same number of events are triggered. For instance, firing on a station with the "stationAI.plist" will trigger 2 alertConditionChanged events, and that will now happen with the "oolite-stationAI.js". 

However, in thinking about this a bit more, I'm wondering if we also need to add the event whenever the "station.alertCondition" property is changed. That would allow OXP's to be notified if some other script changes the property. It would probably be a separate PR, though. Thoughts?

AnotherCommander:
I think it sounds logical to fire the event at station.alertCondition  changes and yes, it can be a separate PR.

PR #312, opened 2018-10-17
New atmosphere shader

AnotherCommander:
This is a new atmosphere shader which uses a Fresnel torus around the planet to produce the effect. Overall it looks better than what we had till now.

Performance seems unchanged according to initial tests and some simplifications have been made in the shader to reduce the amount of calculations done. The fade in and out effect at large distances has been removed because it no longer seems necessary. 

I intend to merge this very soon, so if there are any comments or objections, about now would be a good time to be put forward. :-)

### DIFF
--- a/Resources/Shaders/oolite-default-atmosphere.fragment
+++ b/Resources/Shaders/oolite-default-atmosphere.fragment
@@ -27,25 +27,6 @@
 #define DIFFUSE_LIGHT		(gl_LightSource[1].diffuse.rgb)
 #define AMBIENT_LIGHT		(gl_LightModel.ambient.rgb)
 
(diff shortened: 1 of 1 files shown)


AnotherCommander:
Not 100% sure, but  I think this is how it is supposed to look. You can adjust how strong this effect will be by playing with the kFresnelMixFactor. Try something like 0.15 or even less and It should retreat a bit. I prefer the current 0.25 value though due to the lighting that creates when the planet is sitting right next to the sun, but I am open to suggestions.

phkb:
Thinking about it a bit, I guess what I'm seeing is the light reflected in the atmosphere, not on the surface, which makes some sense. I'll have a play with some numbers and let you know what I find.

PR #314, opened 2018-10-18
Adds commsMessageReceived worldScript event when player.commsMessage …

phkb:
…function is called

### DIFF
--- a/src/Core/Scripting/OOJSPlayer.m
+++ b/src/Core/Scripting/OOJSPlayer.m
@@ -378,6 +378,7 @@ static JSBool PlayerCommsMessage(JSContext *context, uintN argc, jsval *vp)
 	}
 	
 	[UNIVERSE addCommsMessage:message forCount:time];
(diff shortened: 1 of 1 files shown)


phkb:
When I originally mentioned this change (quite some time ago, now!) it was mentioned that because this is a breaking change, some sort of notification would be required. Should that just be done on the forum, or in the changelog, or both? Or somewhere else?

AnotherCommander:
The answer is probably  both, but why is it considered a breaking change? What exactly does it break?

phkb:
Any OXP that implements player.commsMessageReceived will now potentially get a null value for the ship parameter, which could be unexpected and break code.

AnotherCommander:
OK, in that case, and given the message sent to the dev list literally minutes ago, I think that we should keep this PR open and merge it after the release of 1.88.

phkb:
No problem. Keeps it simple.

AnotherCommander:
OK, let's pull in the first post-1.88 request...

PR #318, opened 2019-01-13
Fix Mac builds

JensAyton:
These are the changes required to build with Xcode 10, which will bring back nightly builds. It also updates various project files, breaking compatibility with older versions of Xcode.

Done as a PR because merging this will drop support for Mac OS X 10.6 (the 2009–2010 version).


KonstantinosSykas:
Merging this PR will most probably leave macOS out of future releases, until another maintainer with Xcode10 enabled environment joins the team. 

For the time being, Travis CI compiling AI.m and OOTrumble.m fails, complaining about OOViewID.tbl file not found.
Could it be the version of Xcode used by Travis CI? 

Our .travis.yml file tells Travis CI to use Xcode 7.3 and my understanding is that this PR's changes break compatibility with Xcode versions less than 10.
```
language: objective-c
osx_image: xcode7.3

xcode_project: Oolite.xcodeproj
xcode_scheme: 'Oolite - Test Release'
```
Could you update this PR with a .travis.yml having a compatible (with the changes committed) osx_image value? 
For the list of valid osx_image values you may check the relevant Travis CI reference page [here](https://docs.travis-ci.com/user/reference/osx#macos-version).

phkb:
Not sure why the HeadUpDisplay.m and OOConstToJSString.m files have been included in this PR. They seem to be removing things that are unrelated to an Mac-specific code. Accidental, perhaps?

KonstantinosSykas:
@phkb 
Both fixes deal with compilation warnings on Mac. In Linux and Windows, these warnings did not appear due to different compiler options used.

phkb:
Ah, got it. It's still a little confusing, though, because there doesn't appear to be any difference between the lines being removed and similar ones around them. For instance, why would the Mac have a problem with lines 119-123 of OOConstToJSString.m, when they're almost identical to lines 114-117? Why do those lines compile, but lines 119-123 don't? 

It's also probably worth noting that, by applying the changes to OOConstToJSString.m, commit 835770 would need to be reverted as that change relies on the new table.

maikschulz:
> These are the changes required to build with Xcode 10, which will bring back nightly builds. It also updates various project files, breaking compatibility with older versions of Xcode.
> 
> Done as a PR because merging this will drop support for Mac OS X 10.6 (the 2009–2010 version).

I just tried on a freshly setup Xcode 13.4.1 on MacOS 12.5.1 on a MacBook Air with M1:

the libstdc++ errors are fixed, but now the libpng build fails with the message "error: Build input file cannot be found: '/Users/maikschulz/Documents/projects/oolite/deps/libpng/pngwutil.c' (in target 'libpng' from project 'libpng')", both for pngwutil.c and pngwtran.c

Could you kindly update your PR please? Thank you, -Maik

HiranChaudhuri:
There is likely noone in a position to bring this branch up to speed. This PR is doomed.

PR #320, opened 2019-01-16
Change FAQ reference from PLAYING.TXT to controller wiki

### LINKED ISSUE #319: Missing file in FAQ
The [2nd entry](https://github.com/OoliteProject/oolite/blob/1.88/Doc/FAQ.TXT#L17) in the FAQ recommends using Oolite's PLAYING.TXT to learn more alongside _Ian Bell's Flight Training Manual_. This is misleading because the file is not present in the Git repository.

The PLAYING.TXT file is mentioned in both de60dba and 0c126dc but does not appear to be present in either change.

clintmoyer:
The FAQ in the documentation was referencing a non-existent file PLAYING.TXT to provide controller information specific to the Oolite project. This file is not present in the repository and so we can instead link to the Wiki.

Closes #319


AnotherCommander:
Thanks for catching that. Merging.

PR #326, opened 2019-04-06
16 bits per color component as startup parameter

AnotherCommander:
Trying to find a fix for the color banding issue that we see sometimes on planets on any graphics card that does not do dithering. Seems to work well on Windows, Linux needs some more testing but it seems to work too.

The solution is to request OpenGL to set up a 16 bits per color component buffer on stasrtup. However, since there has been one report of 100% usage on the video card (not reproducible at this time), we set the 16bpcc only as an option, if the user requests it. The option can be toggled from both the command line and the defaults file.


AnotherCommander:
Unless we get reports of issues, this PR is scheduled for merging at 12/04/2019.

PR #328, opened 2019-04-12
Provides JS access to the shipyard of a station

phkb:
Adds station.shipyard property to station to view current ships in shipyard.
Adds station.addShipToShipyard to add new ship to shipyard
Adds station.removeShipFromShipyard to remove a ship from the shipyard.

eg of adding ship:
```
    var def = {
        short_description:"A sample ship to demonstrate adding a ship to the shipyard",
        shipdata_key:"cobra1_player_able",
        price:120000,
        personality:Math.floor(Math.random() * 32768),
        extras:["EQ_FUEL_INJECTION","EQ_GAL_DRIVE"]
    };
    player.ship.dockedStation.addShipToShipyard(def);
```

eg of removing ship
 `   player.ship.dockedStation.removeShipFromShipyard(1); // removes ship at index position 1`

### DIFF
--- a/src/Core/Entities/PlayerEntityContracts.m
+++ b/src/Core/Entities/PlayerEntityContracts.m
@@ -1475,21 +1475,11 @@ static NSMutableDictionary *currentShipyard = nil;
 	}
 	if ([station localShipyard] == nil)
 	{
-		[station setLocalShipyard:[UNIVERSE shipsForSaleForSystem:system_id withTL:stationTechLevel atTime:ship_clock]];
+		[station generateShipyard:stationTechLevel];
 	}
 		
 	NSMutableArray *shipyard = [station localShipyard];
 		
-	// remove ships that the player has already bought
-	for (i = 0; i < [shipyard count]; i++)
-	{
-		NSString *shipID = [[shipyard oo_dictionaryAtIndex:i] oo_stringForKey:SHIPYARD_KEY_ID];
-		if ([shipyard_record objectForKey:shipID])
-		{
-			[shipyard removeObjectAtIndex:i--];
-		}
-	}
-	
 	[currentShipyard release];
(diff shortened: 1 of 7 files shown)


src/Core/Entities/PlayerEntityContracts.m
```diff
@@ -1601,6 +1591,29 @@ - (void) setGuiToShipyardScreen:(NSUInteger)skip
...
+	for (i = 0; i < [shipyard count]; i++)
+	{
+		NSString *shipID = [[shipyard oo_dictionaryAtIndex:i] oo_stringForKey:SHIPYARD_KEY_ID];
+		if ([shipyard_record objectForKey:shipID])
+		{
+			[shipyard removeObjectAtIndex:i--];
+		}
+	}
+}*/
+
```
AnotherCommander: This here should be removed. Best not have commented out code in a PR.

AnotherCommander:
Very nice and should help a lot with testing, too. After a quick browse it looks alright, but I'll need to have a better look at it a bit later and see it in action.

src/Core/Entities/PlayerEntityContracts.m
phkb: Fixed. Thanks!

AnotherCommander:
Tested it, seems to work great. Only suggestion I have is to add this just below the `[shipyard addObject:shipyardDefinition];` and `[shipyard removeObjectAtIndex:shipIndex];` lines in OOJSStation.m:
```
if([PLAYER guiScreen] == GUI_SCREEN_SHIPYARD)
{
	[PLAYER setGuiToShipyardScreen:0];
}
```
This way, if you add or remove a ship from the shipyard and happen to be already in the shipyard screen, the screen will be automatically refreshed rather than having to exit it and re-enter it in order to see the change.

PR #330, opened 2019-04-17
addShipToShipyard look for and add lasers from extras array

phkb:
This is so a laser weapon can be included in "extras" array and added to the ship.

### DIFF
--- a/src/Core/Entities/PlayerEntityContracts.m
+++ b/src/Core/Entities/PlayerEntityContracts.m
@@ -1771,9 +1771,6 @@ static NSMutableDictionary *currentShipyard = nil;
 	OOCreditsQuantity price = [shipInfo oo_unsignedLongLongForKey:SHIPYARD_KEY_PRICE];
 	OOCreditsQuantity tradeIn = [self tradeInValue];
 
-	// make sure these two values are always different
-	if (tradeIn == (price * 10)) tradeIn += 1;
-
 	if (credits + tradeIn < price * 10)
 		return NO;	// you can't afford it!
 	
--- a/src/Core/Entities/PlayerEntityContracts.m
+++ b/src/Core/Entities/PlayerEntityContracts.m
@@ -1771,6 +1771,8 @@ static NSMutableDictionary *currentShipyard = nil;
 	OOCreditsQuantity price = [shipInfo oo_unsignedLongLongForKey:SHIPYARD_KEY_PRICE];
 	OOCreditsQuantity tradeIn = [self tradeInValue];
 
+	// make sure these two values are always different
+	if (tradeIn == (price * 10)) tradeIn += 1;
 
 	if (credits + tradeIn < price * 10)
 		return NO;	// you can't afford it!
--- a/src/Core/Entities/PlayerEntityControls.m
+++ b/src/Core/Entities/PlayerEntityControls.m
@@ -2666,10 +2666,9 @@ static NSTimeInterval	time_last_frame;
 						shipprice = [self priceForShipKey:key];
 					}
 
-					OOCreditsQuantity money = credits;
 					if ([self buySelectedShip])
 					{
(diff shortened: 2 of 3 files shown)


AnotherCommander:
Bug:
1. Spawn at Lave.
2. F3-F3 to go to shipyard. It''s empty.
3. Enter the following in the console:
-`P.credits+=10000000`
-`var def = {
        short_description:"A sample ship to demonstrate adding a ship to the shipyard",
        shipdata_key:"cobramk1-player",
        price:120000,
        personality:Math.floor(Math.random() * 32768),
        extras:["EQ_FUEL_INJECTION","EQ_GAL_DRIVE"]
    };
    player.ship.dockedStation.addShipToShipyard(def);`
-`var def = {
        short_description:"A Milsample ship to demonstrate adding a ship to the shipyard",
        shipdata_key:"cobra3-player",
        price:120000,
        personality:Math.floor(Math.random() * 32768),
        extras:["EQ_WEAPON_MILITARY_LASER","EQ_FUEL_INJECTION","EQ_GAL_DRIVE"]
    };
    player.ship.dockedStation.addShipToShipyard(def);`
-`var def = {
        short_description:"A Miningsample ship to demonstrate adding a ship to the shipyard",
        shipdata_key:"cobra3-player",
        price:120000,
        personality:Math.floor(Math.random() * 32768),
        extras:["EQ_WEAPON_MINING_LASER","EQ_FUEL_INJECTION","EQ_GAL_DRIVE"]
    };
    player.ship.dockedStation.addShipToShipyard(def);`
You now have three ships in the shipyard and enough money to buy any of them.
4. Buy the first Cobra MKIII. All good so far.
5. Return immediately to the shipyard and try to buy the second Cobra. Clock runs, but ship is not removed from shipyard until you manually exit the screen and return and there is no automatic transition to the status screen once you press enter to buy.

phkb:
Oh man! Back to the drawing board...

phkb:
Actually, that doesn't completely work either. Seems to work if the shipyard already has a ship in it, but not if the shipyard is empty.

phkb:
OK, I've tracked down the issue. The problem is that the trade-in price of the 1st Cobra III you bought is exactly the same as the purchase price of the 2nd Cobra III. In that instance, the player's credits ends up exactly the same at the end of the process, and the calling function is looking for a change in credit value to determine whether the purchase happened successfully.

I'll commit a simple fix for this, but I don't know if I like it or not. Feel free to send it back for further work.

AnotherCommander:
Not really sure about this fix either. It seems to circumvent a problem rather than hitting its root cause. 

This code here in PlayerEntityControls.m has the check for change of credits once a ship has been bought in the shipyard:
```
OOCreditsQuantity money = credits;
if ([self buySelectedShip])
{
	if (money != credits)	// money == credits means we skipped to another page, don't do anything
	{
		[UNIVERSE removeDemoShips];
		[self setGuiToStatusScreen];
		[self playBuyShip];
		[self doScriptEvent:OOJSID("playerBoughtNewShip") withArgument:self andArgument:[NSNumber numberWithUnsignedLongLong:shipprice]]; // some equipment.oxp might want to know everything has changed.
	}
}
```
It looks like the check `if (money != credits)` was put there to handle the cases where the player presses Enter on More: or Back: rather than on an actual ship in the list. So to me, it looks like we are hitting an edge case, which most likely was always present but never manifested before.

I wonder if a better check for that would be `if (![key hasPrefix:@"More:"])`, which explicitly checks for the More: and Back: cases. @phkb can you give it a try with this (after reverting your fix) and see if there are any issues when buying ships? I did a very quick test and did not spot anything (multiple ship purchases bug gets fixed too with this), but more testing would be preferred.

phkb:
I did a variety of tests with that IF clause, but couldn't get it to break. I think it looks OK now.

AnotherCommander:
OK, let's do it then.

PR #334, opened 2019-04-26
Main menu revamp: Replacing View Keyboard Settings menu item with Game Options. Game Opt…

AnotherCommander:
…ions items that are controlled by savegames are greyed out when viewed from the intro screen. Some code refactoring. Testing required, especially stick mapping.


AnotherCommander:
Following the forum discussion here: http://www.aegidian.org/bb/viewtopic.php?f=3&p=267090#p267090
I think we are good to go ahead with the merge.

phkb:
I’m good with it!

PR #335, opened 2019-04-29
Adding EquipmentInfo.calculatedPrice property, adding 'paid' param to…

phkb:
… playerBoughtEquipment event

These two changes help with gaining better visibility of equipment prices.

EquipmentInfo.calculatedPrice is the calculated price of the equipment item, if the equipment item utilises the updateEquipmentPrice() condition script function.

Adding the "paid" parameter to the playerBoughtEquipment event gives instant visibility of the actual price paid by the player, taking into account any rebates or adjustments in the purchase.

### DIFF
--- a/src/Core/Entities/PlayerEntity.h
+++ b/src/Core/Entities/PlayerEntity.h
@@ -1000,6 +1000,8 @@ typedef enum
 - (void) setFastEquipmentA:(NSString *)eqKey;
 - (void) setFastEquipmentB:(NSString *)eqKey;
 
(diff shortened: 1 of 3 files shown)


AnotherCommander:
With the extra parameter in playerBoughtEquipment, are we still backwards compatible?

phkb:
@AnotherCommander We should be, and my testing says yes. For existing OXP's that implement playerBoughtEquipment, the new parameter will just be ignored.

PR #336, opened 2019-04-30
Adding playerWillBuyNewShip world event

phkb:
Gives OXP's an opportunity to do something before the player ship is actually sold/replaced.
Event is in two forms: 

1. When player actually buys a ship from the F3F3 shipyard.
  `this.playerWillBuyNewShip = function(shipKey, shipyard_info, price, tradeIn) {  }`

2. When player ship is changed via the JS method player.replaceShip("shipkey")
  `this.playerWillBuyNewShip = function(shipKey) {  }`

### DIFF
--- a/src/Core/Entities/PlayerEntityContracts.h
+++ b/src/Core/Entities/PlayerEntityContracts.h
@@ -116,7 +116,7 @@ MA 02110-1301, USA.
 
 - (OOCreditsQuantity) priceForShipKey:(NSString *)key;
 - (BOOL) buySelectedShip;
-- (BOOL) buyNamedShip:(NSString *)shipName;
+- (BOOL) replaceShipWithNamedShip:(NSString *)shipName;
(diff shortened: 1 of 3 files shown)


AnotherCommander:
Given that when replacing ship no change in credits is applied, should that not be a differently named event (e.g. playerWillReplaceShip)?

phkb:
I think I was just following the standard used for playerBoughtNewShip, which is the same for both methods even though one of them is a replace rather than a buy. Happy to go with your suggestion though.

AnotherCommander:
Since buyNamedShip is called exclusively by PlayerReplaceShip in OOJSPlayer.m, I think it would be best to rename that too to something more akin to what it actually does, e.g. replaceShipWithNamedShip or something.

phkb:
Renamed the function as suggested. I also switched the JS event names for the replaceShip process, making them "playerWillReplaceShip" and "playerReplacedShip". However, I've left the "playerBoughtNewShip" event where it is for the moment, to stay backwards compatible with any OXP that's expecting to see this event for a player.replaceShip.

phkb:
Took out the unnecessary "price" parameter from the "playerReplacedShip" event.

PR #337, opened 2019-05-01
Limiting star and nebula textures by galaxy

phkb:
This adds the ability to limit star and nebula textures to particular galaxies, providing the opportunity to make galaxies more visually unique. 

At the same time, in order to make the merging of star and nebula textures more logical and consistent, they are now being merged on the "texture" key. Therefore, if an OXP overrides the "oolite-nebula-1.png" definition in a "nebulatextures.plist" file (for instance, adding a galaxy or changing the probability), they can be confident this change has been applied. To allow for the same texture to appear in multiple galaxies, an optional "key" property can be added to the dictionary, which will take the place of the "texture" property when merging the arrays.

To demonstrate this, use the following in place of the current nebulatextures.plist, which will allocate one nebula image per galaxy.

```
(
	{
		texture = "oolite-nebula-1.png";
		probability = 1.0;
		galaxy = 0;
	},
	{
		texture = "oolite-nebula-2.png";
		probability = 1.0;
		galaxy = 1;
	},
	{
		texture = "oolite-nebula-3.png";
		probability = 1.0;
		galaxy = 2;
	},
	{
		texture = "oolite-nebula-4.png";
		probability = 1.0;
		galaxy = 3;
	},
	{
		texture = "oolite-nebula-1.png";
		key = "nebula1-gal4";
		probability = 1.0;
		galaxy = 4;
	},
	{
		texture = "oolite-nebula-2.png";
		key = "nebula2-gal5";
		probability = 1.0;
		galaxy = 5;
	},
	{
		texture = "oolite-nebula-3.png";
		key = "nebula3-gal6";
		probability = 1.0;
		galaxy = 6;
	},
	{
		texture = "oolite-nebula-4.png";
		key = "nebula4-gal7";
		probability = 1.0;
		galaxy = 7;
	}
)

```
If no "galaxy" property is set, existing behaviour should be unchanged.

Outstanding: I really need someone to confirm I've done the alloc/dealloc correctly.

### DIFF
--- a/src/Core/OOProbabilisticTextureManager.m
+++ b/src/Core/OOProbabilisticTextureManager.m
@@ -29,6 +29,7 @@ SOFTWARE.
 #import "ResourceManager.h"
 #import "OOTexture.h"
 #import "OOCollectionExtractors.h"
+#import "PlayerEntityScriptMethods.h"
 
 
 @implementation OOProbabilisticTextureManager
@@ -54,12 +55,14 @@ SOFTWARE.
 {
 	BOOL				OK = YES;
 	NSArray				*config = nil;
-	NSUInteger			i, count;
+	NSUInteger			i, count, j;
 	id					entry = nil;
 	NSString			*name = nil;
 	float				probability;
 	OOTexture			*texture = nil;
-	
+	int 				galID = -1;
+	id					object = nil;
+
 	self = [super init];
 	if (self == nil)  OK = NO;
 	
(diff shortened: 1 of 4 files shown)


AnotherCommander:
Really pressed with time these days and can't do proper review. @phkb if you have tested it sufficiently and it does not crash, we can merge it. The only comment I have is the value 8 which is hardcoded. If we have a constant representing number of galaxies somewhere, let's use that.

phkb:
I've tested a raw install (no addons), an install with just my amended nebulatextures.plist above, and with my normal game (300+ OXP's and a customised version of Zygo's Cinematic Skies), and found no crashes with the last tweak. I think this is good to merge.

src/Core/OOProbabilisticTextureManager.m
```diff
@@ -79,14 +79,14 @@ - (id)initWithPListName:(NSString *)plistName
 		_textures = malloc(sizeof *_textures * count);
 		_prob = malloc(sizeof *_prob * count);
 		_galaxy = malloc(sizeof *_galaxy * count);
-		_probMaxGal = malloc(sizeof *_probMaxGal * 8);
+		_probMaxGal = malloc(sizeof *_probMaxGal * kOOMaximumGalaxyID);
```
AnotherCommander: I think this should be `_probMaxGal = malloc(sizeof *_probMaxGal * (kOOMaximumGalaxyID + 1));`

phkb:
I knew that malloc stuff would bite me somewhere...

PR #344, opened 2019-12-18
New nav beacon

phkb:
Utilitises (with permission) Cmd. Cheyd's DH Navigation Buoy to replace the core buoy model, bringing this item into more visual alignment with Griff's detailed textures.


AnotherCommander:
Travis fails to build the project: https://travis-ci.org/OoliteProject/oolite

The problem is that the oolite-binary-resources submodule has not been updated to revision ac1e12. You will need to push the revision changes on your local hard drive to the submodule to fix this. If all has gone well, oolite-binary-resources on GitHub should contain the new nav buoy binary files and show ac1e12 as its last revision.

Before merging the PR, it would be good to also update the documentation. The Reference Sheet nav buoy image will need to be replaced with the new one.

phkb:
Apologies! I think I've got everything in now, and updated docs to match.

AnotherCommander:
Looks good, although I can't physically test it at the moment. However, I would recommend adding a gloss value in shipdata. I was using `gloss = 0.65;` and thought it improved the looks a lot.

phkb:
That's a relief! Excellent. And gloss value is now in play.

PR #355, opened 2020-07-22
Protection for ship launches

phkb:
In case where ship is too big for a dock and can't be launched.


AnotherCommander:
It's fine, but I think the push(ship) line should be enclosed in curly braces to conform to the general coding style used in the file.

PR #356, opened 2020-07-23
Changes shape of icon on basic compass when station is target

phkb:
Makes it clearer what is being targeted.
Code contributed by forum member Milo.


AnotherCommander:
Needs an update to the documentation (OoliteRS.pdf) for completeness.

PR #358, opened 2020-07-27
Check and update of doc using latest LibreOffice

phkb:
After the issue with the PDF doc being messed up, I went back to the process I used and had another look at the odt file. It appears that the original odt file was not opening successfully in the latest LibreOffice (6.4.5.2). This PR corrects the formatting and layout issues, bringing it up to date with this release of LibreOffice, and the exported PDF file has image DPI set at 600 for slightly clearer ship images.

If someone has an earlier version of LibreOffice it would be beneficial to check this PR out and see if everything is still aligned nicely.


AnotherCommander:
Looks fine. Since this PR is already about updating the odt and pdf, can I ask that you update the copyright dates to 2020 in the two documents? We can then merge this.

PR #361, opened 2020-08-11
Update xcconfig settings:

MaddTheSane:
* Quiet new warnings.
* Add arm64 symbol file.


AnotherCommander:
Before merging this PR, do we know if there is a way to silence these messaging id warnings in the new Clang without telling the compiler to not emit them and without breaking the other two ports?

PR #366, opened 2020-08-30
Adds weaponInfo to EquipmentInfo script object

dybal:
weapon_info was included in equipment.plist in v1.81, but there's no way for scripts to recover that information, so I added it to EquipmentInfo scripting object as 'weaponInfo', a dictionary with the weapon_info fields from equipment.plist (used scripInfo as template)

I have been using this mod for 9 days (Towbar v0.105 's Laser Reductor uses it to set the amount of energy to give back when a derelicts is hit), the glitches were all on the script side.

There are a few other OXPs that use weapon_info, but they are expecting a dictionary with name "weapon_info"... the property I added is camel case, like all other scripting object's properties, so those OXPs will have to be updated to use the new property

### DIFF
--- a/src/Core/OOEquipmentType.h
+++ b/src/Core/OOEquipmentType.h
@@ -142,6 +142,7 @@ SOFTWARE.
 // weapon properties
 - (BOOL) isTurretLaser;
 - (BOOL) isMiningLaser;
+- (NSDictionary *) weaponInfo;
 - (GLfloat) weaponRange;
 - (GLfloat) weaponEnergyUse;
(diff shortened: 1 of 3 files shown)


dybal:
I guess we are out of feature freeze... my apologies if we are not.

AnotherCommander:
We are, no worries.

I'll take a look at this after I've had a few days' break - preparing the release has been a quite exhausting task, From a quick glance it seems good, but it needs proper testing. If anyone else wants to review it in the meantime and they decide it's good to merge, feel free to go ahead.

PR #369, opened 2020-09-22
Doesn't block an equipment installation if an incompatible equipment had been installed but has already been removed

dybal:
An equipmentKey isn't removed from oolite-equipment-control's this.$equipmentEnabled dictionary keys when an equipment is removed, the value for that key is changed to 0 if the equipment was removed (and to 1 if it's installed or re-installed).

So not all keys in this.$equipmentEnabled represent equipments that are installed in the ship and it's necessary to check the value for that key in the dictionary to make sure the equipment is installed and only after that look for the equipments incompatible with it.

Without this correction, if an equipment was installed it would block the installation of equipments incompatible with it even after it was uninstalled.

To reproduce, install Smivs Battle Damage OXP and remove the equipment EQ_HULL_REPAIR from a ship and then try to award it back (it's incompatible with self)

### DIFF
--- a/Resources/Scripts/oolite-equipment-control.js
+++ b/Resources/Scripts/oolite-equipment-control.js
@@ -159,9 +159,11 @@ this.equipmentAdded = function(equip)
 	var eq_dict = this.$equipmentEnabled;
 	var eqInfo;
 	for (var eqKey in eq_dict) {
-		eqInfo = EquipmentInfo.infoForKey(eqKey);
-		if (eqInfo && eqInfo.incompatibleEquipment && eqInfo.incompatibleEquipment.indexOf(equip) >= 0) {
-			player.ship.removeEquipment(eqKey);
+		if (this.$equipmentEnabled[eqKey]) {
+			eqInfo = EquipmentInfo.infoForKey(eqKey);
+			if (eqInfo && eqInfo.incompatibleEquipment && eqInfo.incompatibleEquipment.indexOf(equip) >= 0) {
+				player.ship.removeEquipment(eqKey);
+			}
 		}
 	}
 	this.$equipmentEnabled[equip] = 1;
--- a/Resources/Scripts/oolite-equipment-control.js
+++ b/Resources/Scripts/oolite-equipment-control.js
@@ -156,9 +156,11 @@ this.equipmentAdded = function(equip)
 	var eq_dict = this.$equipmentEnabled;
 	var eqInfo;
 	for (var eqKey in eq_dict) {
-		eqInfo = EquipmentInfo.infoForKey(eqKey);
-		if (eqInfo && eqInfo.incompatibleEquipment && eqInfo.incompatibleEquipment.indexOf(equip) >= 0) {
-			player.ship.removeEquipment(eqKey);
+		if (this.$equipmentEnabled[eqKey]) {
+			eqInfo = EquipmentInfo.infoForKey(eqKey);
+			if (eqInfo && eqInfo.incompatibleEquipment && eqInfo.incompatibleEquipment.indexOf(equip) >= 0) {
+				player.ship.removeEquipment(eqKey);
+			}
 		}
 	}
 	this.$equipmentEnabled[equip] = 1;


Resources/Scripts/oolite-equipment-control.js
```diff
@@ -159,9 +159,11 @@ this.equipmentAdded = function(equip)
 	var eq_dict = this.$equipmentEnabled;
 	var eqInfo;
 	for (var eqKey in eq_dict) {
-		eqInfo = EquipmentInfo.infoForKey(eqKey);
-		if (eqInfo && eqInfo.incompatibleEquipment && eqInfo.incompatibleEquipment.indexOf(equip) >= 0) {
-			player.ship.removeEquipment(eqKey);
+		if (this.$equipmentEnabled[eqKey]) {
```
phkb, line 159: Would it be better to use eq_dict here, rather than this.$equipmentEnabled?

mcarans:
@phkb Do you have any idea if this is still a problem? I can see that the code in Resources/Scripts/oolite-equipment-control.js still looks like the code this PR intends to change ie. around line 158:
```
	for (var eqKey in eq_dict) {
		eqInfo = EquipmentInfo.infoForKey(eqKey);
		if (eqInfo && eqInfo.incompatibleEquipment && eqInfo.incompatibleEquipment.indexOf(equip) >= 0) {
			player.ship.removeEquipment(eqKey);
```

It would be easy to apply this change, but I have no idea if this problem was already fixed in some other way: "if an equipment was installed it would block the installation of equipments incompatible with it even after it was uninstalled."

mcarans:
@oocube I tried to test this and installed Smivs Battle Damage OXP, but I'm no expert with the debug console and couldn't figure out how to do "remove the equipment EQ_HULL_REPAIR from a ship and then try to award it back (it's incompatible with self)". Do you know how to do this or if not how to test the general aim of this PR "Doesn't block an equipment installation if an incompatible equipment had been installed but has already been removed"?

phkb:
I think I did an update to Battle Damage that changed the way it worked so that the issue in this PR isn't hit anymore. I'll hopefully have a chance to create a new test for this later today.

mcarans:
@phkb I managed to get this fork to rebase through the GitHub GUI, so it can be easily merged if there is currently a problem with equipment installs being blocked. I was able to start the game on Windows and buy and sell equipment so it hasn't broken anything, but I have not tested the central claim of the PR as I'm not sure how. Please let me know if you need me to assist in testing anything.

phkb:
[TestingRemoval.oxp.zip](https://github.com/user-attachments/files/32137613/TestingRemoval.oxp.zip)
This zip has a OXP that demostrates the problem.

Negative test: In an unpatched version of Oolite, you can buy "Item A" once, and then remove "Item A", but "Item A" will not reappear in the available equipment list.

In a patched version of Oolite, you can buy "Item A" once, then remove "Item A", and do it as many times as you like. Item A is not being blocked from installation.

phkb (approved):

Resources/Scripts/oolite-equipment-control.js
phkb: I think my initial examination of the code wasn't helpful. It can be ignored.

oocube:
> @oocube I tried to test this and installed Smivs Battle Damage OXP, but I'm no expert with the debug console and couldn't figure out how to do "remove the equipment EQ_HULL_REPAIR from a ship and then try to award it back (it's incompatible with self)". Do you know how to do this or if not how to test the general aim of this PR "Doesn't block an equipment installation if an incompatible equipment had been installed but has already been removed"?

I'm no expert on the debug console either. I implemented the debug console protocol but for different purposes so I'd need to dig out how your use case can be solved using either Nexus (my first stab at this) and later Oolitestarter.. Yet MrFlibble improved the python based debug console which might be more straightforward in this case.

For the exact commands to run I would not know either.

phkb:
I'm pretty sure I updated Battle Damage to avoid running into this issue, so you probably can't use it as an example. Which is why I added a sample OXP to the comments. You don't need the debug console to test it. It works purely with in game functions.

mcarans:
> [TestingRemoval.oxp.zip](https://github.com/user-attachments/files/32137613/TestingRemoval.oxp.zip) This zip has a OXP that demostrates the problem.
> 
> Negative test: In an unpatched version of Oolite, you can buy "Item A" once, and then remove "Item A", but "Item A" will not reappear in the available equipment list.
> 
> In a patched version of Oolite, you can buy "Item A" once, then remove "Item A", and do it as many times as you like. Item A is not being blocked from installation.

That worked for me. Thanks!

mcarans:
@dybal Merging this PR. Thanks for your contribution!

PR #373, opened 2020-10-25
Fix an error reported by gcc about format-security

bcoconni:
On Fedora 32, gcc 10.2.1 reports an error:
```bash
 Compiling file src/SDL/MyOpenGLView.m ...
src/SDL/MyOpenGLView.m: In function ‘-[MyOpenGLView init]’:
src/SDL/MyOpenGLView.m:327:2: error: format not a string literal and no format arguments [-Werror=format-security]
  327 |  OOLog(@"display.initGL", @"Achieved color / depth buffer sizes (bits):");
      |  ^
cc1obj: some warnings being treated as errors
```
This PR fixes the issue on my system.

### DIFF
--- a/src/SDL/MyOpenGLView.m
+++ b/src/SDL/MyOpenGLView.m
@@ -324,7 +324,7 @@ MA 02110-1301, USA.
 	}
 	
 	int testAttrib = -1;
-	OOLog(@"display.initGL", @"Achieved color / depth buffer sizes (bits):");
+	OOLog(@"display.initGL", @"%@", @"Achieved color / depth buffer sizes (bits):");
 	SDL_GL_GetAttribute(SDL_GL_RED_SIZE, &testAttrib);
 	OOLog(@"display.initGL", @"Red: %d", testAttrib);
 	SDL_GL_GetAttribute(SDL_GL_GREEN_SIZE, &testAttrib);


AnotherCommander:
Begin rant: Why on Earth is this treated as an error on Fedora? It is perfectly legal Obj-C syntax and should at the worst case be flagged as a warning. The proposed fix works of course (and it is not the first tine we've had to apply something like this either) but the code is more confusing and less readable than before after it has been applied. And, just to be clear, I am not trying to be harsh to anyone, just trying to understand why some versions of gcc seem to be so extremely pedantic, to the point of refusing legal syntax.

Another fix that will work perfectly fine and without any security risk would be to set the -Wno-error=format-security flag somwhere in GNUmakefile.

End rant. Phew. :-)

Since we have already applied similar things in the past, this PR will mist likely make it to merge. Just to be sure that there will be no more of those though, was that the only occurrence? Does the build succeed after you have corrected it?

bcoconni:
> Another fix that will work perfectly fine and without any security risk would be to set the -Wno-error=format-security flag somewhere in GNUmakefile.

Yes, but I don't feel comfortable disabling security related flags even when they seem pedantic. Furthermore, a similar issue has already been discussed in the PR #198 but the subject has been dropped at the time. However the changes have eventually found their way to the code (see commit 0b12ed3a228f0e4 for example). So, long story short, I have submitted this PR.
> Since we have already applied similar things in the past, this PR will most likely make it to merge. Just to be sure that there will be no more of those though, was that the only occurrence? Does the build succeed after you have corrected it?

Yes, the build succeeds once the attached patch is applied and `oolite` runs smoothly 😄

PR #375, opened 2020-10-29
Unlock Game Load/Save screen when paused while docked.

KonstantinosSykas:
Currently if you pause when in Game Load/Save screen and while docked the controls are locked and no options can be altered. On the other hand if you pause while in flight and activate the Game Load/Save screen then options are available for configuration.
This change is to ensure Game Load/Save screen consistent behavior when paused while docked and when paused while in flight.


KonstantinosSykas:
Further testing brought up side-effects making obvious that this needs more changes risking other functionality. Therefore, I close this pull request and may return on this matter in the future.

PR #377, opened 2020-12-16
Script assing to weapon mount fix

dybal:
This PR fixes a bug when a weapon is assigned to a weaponMount by script: if the weapon has a condition_script with allowAwardEquipment defined, it is being called with context "purchase" instead of "scripted". (ship.awardEquipment, which is used to award everything but weapons, already sets the context to "scripted")

That breaks Ship Storage Helper OXP when it tries to restore a ship that has a laser with purchase restrictions, such as Weapon Laws OXP creates.

To replicate the bug, install Weapon Laws OXP and try to assign EQ_WEAPON_MILITARY_LASER to a ship at Lave with the Debug Console. The Military laser will not be available in the Ship Outfitting, due to the condition_scripts' allowAwardEquipment allowing purchase of it only in Anarchies, Feudal States and Multi-government systems, but it allows the award in context scripted, npc, etc., so it should be possible for a script to assign it.

### DIFF
--- a/src/Core/Scripting/OOJSShip.m
+++ b/src/Core/Scripting/OOJSShip.m
@@ -1862,7 +1862,11 @@ static JSBool ShipSetProperty(JSContext *context, JSObject *this, jsid propID, J
 					facing = [entity currentWeaponFacing];
 					break;
 			}
-			[entity setWeaponMount:facing toWeapon:sValue];
+			if (entity == PLAYER) {
+				[entity setWeaponMount:facing toWeapon:sValue inContext:@"scripted"];
+			} else {
+				[entity setWeaponMount:facing toWeapon:sValue];
+			}
 			return YES;
 
 		case kShip_maxEscorts:


AnotherCommander:
I am receiving this when attempting to compile the PR: 
``` 
Compiling file src/Core/Scripting/OOJSShip.m ...
src/Core/Scripting/OOJSShip.m: In function 'ShipSetProperty':
src/Core/Scripting/OOJSShip.m:1868:5: warning: 'ShipEntity' may not respond to '-setWeaponMount:toWeapon:inContext:' [enabled by default]
src/Core/Scripting/OOJSShip.m:1868:5: warning: (Messages without a matching method signature [enabled by default]
src/Core/Scripting/OOJSShip.m:1868:5: warning: will be assumed to return 'id' and accept [enabled by default]
src/Core/Scripting/OOJSShip.m:1868:5: warning: '...' as arguments.) [enabled by default]
```

This is one of the serious types of warnings and requires fixing before merging.

src/Core/Scripting/OOJSShip.m
```diff
@@ -1862,7 +1862,11 @@ static JSBool ShipSetProperty(JSContext *context, JSObject *this, jsid propID, J
 					facing = [entity currentWeaponFacing];
 					break;
 			}
-			[entity setWeaponMount:facing toWeapon:sValue];
+			if (entity == PLAYER) {
```
AnotherCommander, line 1865: The standard way of checking for player (avoiding use of a global variable and used throughout the best part of the codebase) is `if ([entity isPlayer])`

phkb:
Overridden by #387

PR #378, opened 2020-12-16
Derelict desctruction is not a kill

dybal:
This PR fixes a scenario where the player can increase his rank by searching derelicts left from NPC's conflicts and destroying them.

It increases the player's score when a NPC chooses (as implemented at the core game level, at ShipEntity's takeEnergyDamage) to abandon ship when its energy is too low after a hit from a player's attack, which has the added benefit of allowing players with moral restrictions to killing to improve their ranking (which is necessary for progress in many of the core game careers).


mwerle:
Didn't original Elite also give kills for Asteroids? I know you at least got a small bounty payment - justification being that you were clearing the space-lanes of dangerous debris..

keaaw:
Noooo my secret strategy to reach Elite has been foiled (if this is ever merged)!

oocube:
This PR is debatable (some users want it, other's do not). And the same functionality can be achieved via OXP as demonstrated in thread https://bb.oolite.space/viewtopic.php?t=20825.

Hence I will drop the branch.

PR #380, opened 2021-02-19
Adding "setShipInfoForKey"

tsoj:
This should add the javascript function `function setShipDataForKey(datakey : String, newShipData : Object)` that sets the shipdata for a specific datakey. This will allow OXPs to fiddle with the properties of all ships directly. For example, an OXP that sets the `shipdata.max_yaw` value to 60% that of `shipdata.max_pitch`, emulating a similar mechanic in Elite: Dangerous, where the maximum yaw is smaller than the pitch. Or an OXP that re-evaluates the prices of ships in the shipyard.

I also added some helpful tips for compiling on Linux.

This is a draft because I am not sure, if I introduced memory leaks, as I don't know Objective-C very well.

### DIFF
--- a/src/Core/OOShipRegistry.m
+++ b/src/Core/OOShipRegistry.m
@@ -209,6 +209,12 @@ static NSString * const	kVisualEffectDataCacheKey = @"visual effect data";
 	return [_shipData objectForKey:key];
 }
 
+- (void) setShipInfoForKey:(NSString *)key with:(NSDictionary *)newShipData
+{
+	NSMutableDictionary *mutableDict = [_shipData mutableCopy];
+	[mutableDict setObject:OODeepCopy(newShipData) forKey:key];
+	_shipData = [mutableDict mutableCopy];
+}
 
 - (NSDictionary *) effectInfoForKey:(NSString *)key
 {
--- a/src/Core/OOShipRegistry.m
+++ b/src/Core/OOShipRegistry.m
@@ -211,7 +211,7 @@ static NSString * const	kVisualEffectDataCacheKey = @"visual effect data";
 
 - (void) setShipInfoForKey:(NSString *)key with:(NSDictionary *)newShipData
 {
-	NSMutableDictionary *mutableDict = [_shipData mutableCopy];
+	NSMutableDictionary *mutableDict = [NSMutableDictionary dictionaryWithDictionary:_shipData];
 	[mutableDict setObject:OODeepCopy(newShipData) forKey:key];
 	_shipData = [mutableDict mutableCopy];
 }
--- a/src/Core/OOShipRegistry.m
+++ b/src/Core/OOShipRegistry.m
@@ -213,7 +213,8 @@ static NSString * const	kVisualEffectDataCacheKey = @"visual effect data";
 {
 	NSMutableDictionary *mutableDict = [NSMutableDictionary dictionaryWithDictionary:_shipData];
 	[mutableDict setObject:OODeepCopy(newShipData) forKey:key];
-	_shipData = [mutableDict mutableCopy];
+	DESTROY(_shipData);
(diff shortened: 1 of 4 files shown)


src/Core/OOShipRegistry.m
```diff
@@ -209,6 +209,12 @@ - (NSDictionary *) shipInfoForKey:(NSString *)key
 	return [_shipData objectForKey:key];
 }
 
+- (void) setShipInfoForKey:(NSString *)key with:(NSDictionary *)newShipData
+{
+	NSMutableDictionary *mutableDict = [_shipData mutableCopy];
+	[mutableDict setObject:OODeepCopy(newShipData) forKey:key];
+	_shipData = [mutableDict mutableCopy];
+}
```
AnotherCommander, line 217: I think you need to release mutableDict after being done using it. Can't check myself right now so I will have to ask, can you try `NSMutableDictionary *mutableDict = [NSMutableDictionary dictionaryWithDictionary:_shipdata];`? I think in this case you will not need to release manually.

The above may not work, but if you can give it a try and it does work, I think it may be a slightly better way of handling it. The general rule is that anything created with a copy selector requires a manual release.

AnotherCommander: Also, _shipData is a dictionary, Why is it assigned a mutableDictionary type at the last line of the method?

tsoj: Your suggestion worked.

I pretty much copy-pasted it from https://stackoverflow.com/a/29391936/6860744

AnotherCommander: In order to avoid further memory leakage, I would recommend changing
```_shipData = [mutableDict mutableCopy];``` (points _shipData to a new memory location without releasing the previous one)
to
```DESTROY(_shipData);``` (releases memory held by _shipData)
```_shipData = [[NSDictionary dictionaryWithDictionary:mutableDict] retain];``` (Reassigns _shipData to new location and retains it so it won't be lost when the method exits).

I have built the PR to test it on my system and, although the change in _shipData clearly takes place, I am unable to see the results of the change in the game. This is probably because there is no instruction to reload the registry with the new changes anywhere. But it seems to me, going through OOShipRegistry.m, that any attempt to reload the ship information would cause a reset of the current _shipData before reading the information back from the game cache, so I am not really sure how this is supposed to practically work. @tsoj: Do you happen to have an example of an OXP actually using the changes made to _shipData as a result of this PR?

Comment (AnotherCommander) [reply src/Core/OOShipRegistry.m:217]: Nevermind the above, I got it to work (silly me, should have realized better at which points in the game the registry values are meant to be read). The changes in the registry do work, so I think that with the proposed changes applied, we are more or less there.

File: src/Core/Scripting/OOJSShip.m [hunk 2]
```diff
@@ -4357,3 +4359,23 @@ static JSBool ShipStaticShipDataForKey(JSContext *context, uintN argc, jsval *vp
...
+	if (argc >= 2)
+	{
+		NSString *key = OOStringFromJSValue(context, OOJS_ARGV[0]);
+		NSDictionary *newShipData = OOJSNativeObjectFromJSObject(context, JSVAL_TO_OBJECT(OOJS_ARGV[1]));
+		[registry setShipInfoForKey:key with:newShipData];
+		OOJS_RETURN_BOOL(YES);
+	}
+	else
+	{
+		OOJSReportBadArguments(context, @"Ship", @"setShipInfoForKey", MIN(argc, 2U), OOJS_ARGV, nil, @"ship role");
```
AnotherCommander, line 4377: This is probably meant to read: ```OOJSReportBadArguments(context, @"Ship", @"setShipInfoForKey", MIN(argc, 2U), OOJS_ARGV, nil, @"ship key");```

Issue #341, opened 2019-07-04
feature request: add update all option to OXP manager


gilbertohasnofb:
It would be really handy to have an `update all` option in the in-game OXP manager which would update all out-of-date expansions. It is really time consuming to update one by one, particularly when one hasn't fired up the game in a while.

PR #381, opened 2021-02-21
Add update all option to OXP manager

tsoj:
Implements #341.

~In some scenarios, it is necessary to "update installed expansion packs" multiple times. For example, if an installed OXP doesn't have the required dependencies already installed at the time of startup, it won't be installed in the first run of "update all". (Can be reproduced if one manually decreases the version number in the manifest of the local OXZ of FeudalStates and Planetfall).  
But I think that these situations shouldn't even happen in the first place, so this shouldn't be a big problem.~ I fixed this with the third commit.

~Unfortunately, I messed something up with GIT so there are some merge conflicts, I'll just create this pull request and then I'll see if I can fix it somehow.~ I believe I solved it.


AnotherCommander:
Looks good here and compiles without any warnings. @tsoj can you please comfirm that this works as intended with a) no installed OXZs, b) one installed OXZ only and c) > 1 installed OXZs?

The only comment I have on the code is that, for consistency with the rest of the codebase, methods should be separated by two newlines, rather than just one. It is a sign of a good PR when the only remark is about newline usage, innit? ;-).

tsoj:
Okay, I've found an issue with updating OXZs that can be updated but also have dependencies that need updating. I fixed the problem by marking OXPs that are `OXZ_INSTALLABLE_UPDATE` but also is  `OXZ_INSTALLABLE_DEPENDENCIES` (meaning that the new, not yet installed, version has unmet dependencies) as `OXZ_INSTALLABLE_UPDATE`.  
(I also fixed a small issue where a more recent local version would be color coded as updatable version when listing OXZs with "List installed expansion packs", instead of dark blue, like when listing OXZs with "Install expansion pack")

I followed your suggestion and tried a bunch of scenarios, in the end they worked out well:

1. With no installed OXZs:
	+ After hitting enter when "Update installed expansion packs" is selected, it immediately switches back to the starting selection ("Install expansion packs"), so the same as if I enter on "List installed expansion packs" after I entered it already.
2. With no updatable OXZs of 1 OXZ:
	- Same as 1.
3. With 1 updatable OXZs of 1 OXZ:
	- After hitting enter it downloads the one updatable OXZ and the goes back to the selection menu.
4. With no updatable OXZs of many OXZ:
	- Same as 1.
5. With 1 updatable OXZs of many OXZ:
	- Same as 3.
6. With many updatable OXZs of many OXZ:
	- Installs all updatable OXZs, i.e. after the update no expansion packs are cyan and there are still the same OXZs installed.
7. With updatable OXZ that also depends on the update of another OXZ:
	- Same as 6.
8. With local OXZ that is more recent than the remote version:
	- The more recent local OXZ doesn't get touched, otherwise same as 6.

Also fixed the newline :)  
Wrote quite a bit of text here for such a small commit :P

AnotherCommander:
Good job. Let's do this.

PR #385, opened 2021-04-13
Formatting on AI.m

CMDR-Piboy314:
### Formatting on AI.m


KonstantinosSykas (commented):
@AnotherCommander it seems that we have a missing pointer fix at line 920?!?

AnotherCommander:
Actually, I think line 120 is a mistake in the PR. The id is already a pointer and the PR makes it an array of pointers, which is not what we want here.

CMDR-Piboy314:
Oops, I added that part by accident!

PR #387, opened 2021-07-23
Fix for correct context when adding weapons to player ship

phkb:
Adding a weapon to the player ship via a script passes a "purchase" context through to the "allowAwardEquipment" condition script call, even when the weapon is added via script. This code change enforces the "scripted" context when a weapon change is made by script.


phkb:
Belatedly realised that this is what dybal was seeking to achieve with #377. No compile errors on this one, though.

PR #395, opened 2021-12-14
Added setVolume to JavaScript interface

tsoj:
[Relevant post on the forum](http://www.aegidian.org/bb/viewtopic.php?p=280941#p280941).

Here is a short exemplary OXP that makes use of this new function:
```javascript
"use strict";

this._volume = 1.0;
this._addition = -0.05;
this._timer = null;

this._changeVolume = function() {
	this._volume += this._addition;

	if(this._volume >= 1.0)
		this._addition = -0.05;
	if(this._volume <= 0.0)
		this._addition = 0.05;
	
	this._volume = Math.min(Math.max(this._volume, 0.0), 1.0);
	
	Sound.setVolume(this._volume); // sets the volume of the currently playing music
}


this.shipLaunchedFromStation = function(stationLaunchedFrom)
{
	Sound.playMusic("yourmusicfile.ogg", false, 1.0);
	this._timer = new Timer(this, this._changeVolume.bind(this), 5, 0.1);
}
```
It starts a song when undocking, and then fades in and fades out again.

### DIFF
--- a/src/Core/OOMusicController.m
+++ b/src/Core/OOMusicController.m
@@ -166,6 +166,12 @@ enum
 }
 
 
+- (void) setMusicGain:(float)gain
+{
+	[_current setMusicGain:OOClamp_0_1_f(gain)];;
+}
+
+
 // Stop without switching iTunes to in-flight music.
 - (void) justStop
 {
--- a/src/Core/OOMusicController.h
+++ b/src/Core/OOMusicController.h
@@ -69,6 +69,8 @@ typedef enum
 - (void) setMissionMusic:(NSString *)missionMusicName;
 - (void) playMissionMusic;
 
+- (void) setMusicGain:(float)gain;
+
 - (void) justStop;
 - (void) stop;
 - (void) stopMusicNamed:(NSString *)name;	// Stop only if name == playingMusic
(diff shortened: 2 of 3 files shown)


src/Core/OOMusicController.m
```diff
@@ -166,6 +166,12 @@ - (void) playMissionMusic
 }
 
 
+- (void) setMusicGain:(float)gain
+{
+	[_current setMusicGain:gain];
```
AnotherCommander, line 171: Bounds checking? I believe this could be something like `[_current setMusicGain:OOClamp_0_1_f(gain)]`

tsoj: Oh yeah, makes sense, fixed it now.

AnotherCommander:
The existing trunk code already allows for sound volume control of individual sound sources. The earlier OXP example should be able to work in trunk if you substitute the line `Sound.setVolume(this._volume); ` with the line `Sound.musicSoundSource().volume = this._volume;` Please see OOJSSoundSource.m. The volume property is read/write.

Which means, that this PR may not be necessary after all. Sorry I didn't catch this earlier, clearly I don't remember the details of the source anymore. Please test and let us know if the PR does something additional to what is already there.

tsoj:
I completely missed that ...

Works fine as far as I can see.

PR #396, opened 2021-12-22
Update oolite.desktop

HiranChaudhuri:
This desktop file added an icon to the desktop manager. But as soon as the application was started it was not recognized as this icon being active, and another icon would get displayed for the active window. With this patch the desktop manager can correctly identify the window and just mark the existing icon as being active.


tsoj:
I can confirm that this has the intended effect for me (I am using Manjaro Gnome) when adding the `StartupWMClass` line to my existing deskop file.

PR #398, opened 2022-01-06
Activated Github Actions

HiranChaudhuri:
We have now Github Actions that automatically execute on every push and that execute on the three target operating systems Linux, MacOS and Windows.

The scripts perform nothing or fail - we still need to refine the correct build. But here are the necessary stubs.


.github/workflows/build-windows.yml
```diff
@@ -25,7 +25,7 @@ jobs:
 
       - name: compile
         run: |
-          $(env:GITHUB_WORKSPACE)\DevelopmentEnvironment\gcc\Msys_x2\1.0\msys.bat'
+          Invoke-Command -FilePath '$(env:GITHUB_WORKSPACE)\DevelopmentEnvironment\gcc\Msys_x2\1.0\msys.bat'
```
AnotherCommander, line 28: If you want to run msys.bat, you must use this syntax:
`& "$env:GITHUB_WORKSPACE\DevelopmentEnvironment\gcc\Msys_x2\1.0\msys.bat"`

HiranChaudhuri:
I agree to the observation and must say that right now I have no clue. Let's find out.

HiranChaudhuri:
for now it does not make sense to pursue this PR

PR #409, opened 2022-10-13
Like ship accumulates shipyard entries

dybal:


dybal:
created by mistake to the parent project

PR #413, opened 2022-11-21
Better organized README.md for compiling on Linux

tsoj:
It took me quite some time to compile on Fedora Linux, so I thought it might be a good idea to add all the information needed to do this to the README.md.


AnotherCommander:
This is fine, just one thing I noticed while going through the updated readme: it still has a reference to SDL_Mixer. Is SDL_Mixer currently being used for anything in Linux? It is certainly not used in Windows since the days of introducing OpenAL. Maybe we can get rid of the readme part mentioning it while we are in the process of reviewing?

tsoj:
I tried compiling without it being installed, and it worked, so I removed the SDL_mixer parts.

AnotherCommander:
OK, cool. Merging.

PR #421, opened 2023-04-14
add Ubuntu build

HiranChaudhuri:
We can now build on Ubuntu again.
For changes on master a release is created, for changes on other branches a prerelease is created. 
This will allow to use the CI/CD pipeline for both releases and for development.

### DIFF
--- a/.github/workflows/build-all.yml
+++ b/.github/workflows/build-all.yml
@@ -1,0 +1,161 @@
+name: build-all
+
+on:
+  workflow_dispatch:
+  push:
+
+jobs:
+  build-linux:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Environment Variables
+        run: |
+          printenv | sort
+
+      - name: install all the necessary packages
+        run: |
+          sudo apt update
+          sudo apt -y install git gobjc gnustep-devel make libsdl1.2-dev libvorbis-dev libopenal-dev g++ libespeak-dev libnspr4-dev
+
+      - name: Checkout Oolite
+        uses: actions/checkout@v3
+        with:
+          path: oolite
+          fetch-depth: 0
+          submodules: true
+
+      - name: show filesystem before build
+        run: |
+          find . -not -path "./oolite/deps/Windows-deps/*" -not -path "./oolite/Mac-specific/*" -not -path "./oolite/.git/*"
+          
+#      - name: this is necessary, as otherwise textures aren't loaded correctly
+#        run: |
+#          rm oolite/deps/Linux-deps/include/png.h
+#          rm oolite/deps/Linux-deps/include/pngconf.h
+
+      - name: compiling
+        run: |
+          cd oolite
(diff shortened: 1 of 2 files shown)


tsoj:
What about removing the out-commented lines? Personally, when I read old code that I haven't written myself, I am often confused by commented out code, as I am not sure if it is actually code that is still important or not.

AnotherCommander:
Looking at the PR, it looks OK but I am not entirely sure about the pre-release part.

If I understand correctly the yml, when a push is made to master an actual full release will be created and any other push will result in a pre-release.. This means that 1.90 will no longer be listed as the latest official release, instead each nightly one will become the latest official. Essentially this means that a rolling release model is adopted. Am I getting it right and, if yes, is this what we want? Right now the nightlies are published as pre-releases and this preserves 1.90 as being the latest official build.

My preference would be to keep 1.90 as latest and maintain the nightlies originating from pushes to master as pre-releases. I think that the way it works now is fine, because the non-master pushes are being tested (i.e. compiled) anyway, it's just that no binaries are published for them. I think that the milestone releases should be the only ones that get the green "official release" tag.

HiranChaudhuri:
> What about removing the out-commented lines? Personally, when I read old code that I haven't written myself, I am often confused by commented out code, as I am not sure if it is actually code that is still important or not (except for stuff that got notes like "# this is for debug only; it creates huge logs and takes a long time to execute", where it is clear why it is commented out).
> 
> Otherwise I don't see anything wrong, and if it works, it works :D

Ok, I will look into either removing the commented out code or add a comment why it is still in there.

HiranChaudhuri:
> Looking at the PR, it looks OK but I am not entirely sure about the pre-release part.
> 
> If I understand correctly the yml, when a push is made to master an actual full release will be created and any other push will result in a pre-release.. This means that 1.90 will no longer be listed as the latest official release, instead each nightly one will become the latest official. Essentially this means that a rolling release model is adopted. Am I getting it right and, if yes, is this what we want? Right now the nightlies are published as pre-releases and this preserves 1.90 as being the latest official build.
> 
> My preference would be to keep 1.90 as latest and maintain the nightlies originating from pushes to master as pre-releases. I think that the way it works now is fine, because the non-master pushes are being tested (i.e. compiled) anyway, it's just that no binaries are published for them. I think that the milestone releases should be the only ones that get the green "official release" tag.

I agree the official releases shall be official releases, I still would like to see intermediate (or call them pre-releases) also for feature branches. that way it would be easier for others to see the effect of changes - especially when having to decide about a merge.
But as I am not a full master of Github Actions or the release system I would have to see the effects before I fully know if my workflow matches what I intended to. Plus I believe there is room for improvement. For example, I am not yet 100% happy with the naming of the Linux build. And we might want to change the labeling, which is currently done on date but I just was not sure how the version numbers get counted in this project.

HiranChaudhuri:
How did this get merged? Sorry, this was not my intention.

PR #435, opened 2023-07-24
Map keys to oxp equipment

phkb:


src/Core/Entities/PlayerEntityKeyMapper.m
```diff
@@ -880,6 +1004,20 @@ - (NSArray *)keyFunctionList
 	[funcList addObject:[self makeKeyGuiDict:DESC(@"oolite-keydesc-key_debug_shaders") keyDef:@"key_debug_shaders"]];
 	[funcList addObject:[self makeKeyGuiDict:DESC(@"oolite-keydesc-key_debug_off") keyDef:@"key_debug_off"]];
 
+	OOLog(@"testing", @"checking count %d", [customEquipActivation count]);
```
AnotherCommander, line 1007: This should not be in the final submission, it is just test code. Similar testing log statements elsewhere should be removed as well before merging.

phkb: Ah! Missed a couple. Thanks!

PR #436, opened 2023-08-04
Adding beginGalacticHyperspaceCountdown to playership

phkb:


src/Core/Entities/PlayerEntity.m
```diff
@@ -7436,6 +7436,13 @@ - (void) setJumpType:(BOOL)isGalacticJump
 	}
 }
 
+
+- (BOOL) galacticJumpType
```
AnotherCommander, line 7440: Is this used anywhere? If not, let's remove it.

PR #442, opened 2023-08-17
Extra gui screen keys

phkb:
One of the probelms that has developed over time is the confusing list of F4 Interface screens that are added by OXP's. It doesn't take long for the list to become long and unwieldy.

The problem is that the F4 page is the only place to put a mission screen page. Even if the mission page is specifically related to the market, or the chart, or the system info view, they all have to go on the F4 page with some basic grouping as the only way to arrange things.

Coupled with this is the usefulness some of the pages have beyond the station. It makes sense to be able to view chart related info, or market info, as provided by an OXP, while in flight. The message has always been "do not show mission screens in flight," but that logic is counter to the fact that we can view the status screen, the manifest screen, do all sorts of interactions on the chart screen, view the market info page, all while in flight.

This PR aims to address both these issues. First, a new global function has been added, "setExtraGuiScreenKeys", which allows key functions to be registered on a specific gui screen. So, for example, the "Ctrl-G" key combination could be registered on the F6 chart screen. When the key combination is pressed on that screen, the callback function is called, similar to how the callback function is called for an F4 interfaces page. Each key combination defined in the properties is given a reference key, which will be passed to the callback function to identify what key combination was pressed.

By being able to link a mission screen to a specific GUI screen, some of the clutter on the F4 screen can be reduced.

Secondly, the warning message about showing mission screens in flight has been removed. Along with this, the "allowInterrupt" flag is defaulted to YES in all cases when in flight, which prevents a mission screen from locking the player out of their viewscreen. Ship models are already ignored in flight, and the exit screen will default to the main view.

The following is an example of setting up a new definition:
```
setExtraGuiScreenKeys(this.name, {
    guiScreen:"GUI_SCREEN_MANIFEST",
    registerKeys:{"key1":[{key:"g",mod1:true}]},
    callback:this.$myGuiScreenCallback.bind(this)
});
```
The format of the method is similar to the station.setInterface method. The key ("this.name" in the above example) can be registered once per GUI_SCREEN. Multiple key combinations can be specified in the "registerKeys" property. Using the same key and GUI_SCREEN will overwrite the previous settings.

There is also a method to remove the extra key definition:
```
clearExtraGuiScreenKeys(this.name, "GUI_SCREEN_MANIFEST");
```

This small OXP (an update to the RegisterKeysDemo1 used previously), demonstrates implementing the system: [RegisterKeysDemo1_v2.oxp.zip](https://app.box.com/s/z0y9x6f0y22go2o12o0z25oelkmmqbwc)


phkb:
I’m planning on merging this weekend, unless there’s concerns/issues. Happy to hold off, too.

PR #446, opened 2023-08-31
Fix for possible memory leak on Mac

phkb:
This is a fix for a possible memory leak on the Mac based on the info here: [https://blog.vucica.net/2011/06/avoiding-memory-leak-in-openal-and-crash-in-openal-for-mac.html](url)

I've been running this change in my local build (Windows) for a while without impact. But it would be good to hear from others (on Linux particularly) to make sure there is no obvious issues there before merging.

It may not be a fix for the issue with sounds on Mac, but it might help.


AnotherCommander:
This seems to be working without any side effects for me.

PR #447, opened 2023-09-07
fix for issue with not being able to use numbers on text entry missio…

phkb:
…n screens


phkb:
For clarity, this issue might only happen if "allowInterrupt" is set to true in the mission screen options.

PR #449, opened 2023-09-13
Fixes issue where dest ID not being passed to shipWillEnterWitchspace…

phkb:
… when cause = carried
(Should have been "missing" not "mission" in branch name. Not hugely important. But only noticed after pushing).

### DIFF
--- a/src/Core/Universe.m
+++ b/src/Core/Universe.m
@@ -1033,7 +1033,7 @@ static GLfloat	docked_light_specular[4]	= { DOCKED_ILLUM_LEVEL, DOCKED_ILLUM_LEV
 		[player setWormhole:wormhole];
 		[player addScannedWormhole:wormhole];
 
-		ShipScriptEventNoCx(player, "shipWillEnterWitchspace", OOJSSTR("carried"));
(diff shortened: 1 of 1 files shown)


src/Core/Universe.m
```diff
@@ -1033,7 +1033,7 @@ - (void) carryPlayerOn:(StationEntity*)carrier inWormhole:(WormholeEntity*)wormh
 		[player setWormhole:wormhole];
 		[player addScannedWormhole:wormhole];
 
-		ShipScriptEventNoCx(player, "shipWillEnterWitchspace", OOJSSTR("carried"));
+		ShipScriptEventNoCx(player, "shipWillEnterWitchspace", OOJSSTR("carried"), INT_TO_JSVAL([wormhole destination]));
```
AnotherCommander, line 1036: Isn't this the same as the somewhat simpler
```
ShipScriptEventNoCx(player, "shipWillEnterWitchspace", OOJSSTR("carried"), INT_TO_JSVAL(dest));
```
dest is set as [wormhole destination] a few lines earlier.

phkb: Yep, good call.

AnotherCommander:
Looks good to merge.

PR #451, opened 2023-10-03
Expands ship.canAwardEquipment for other contexts

phkb:
By default, the "canAwardEquipment" checks the equipment with the "scripted" context. This PR allows other contexts to be tested on an equipment item. 

For example:
`player.ship.canAwardEquipment("EQ_NAVAL_ENERGY_UNIT")` will return true. This is because, with the "scripted" context, it is allowed to be awarded to the player. However, there are restrictions on whether you can purchase the equipment or not. With this PR, you can do the following:
`player.ship.canAwardEquipment("EQ_NAVAL_ENERGY_UNIT", "purchase")`, which will return false if the player has not yet completed the mission it relates to.


AnotherCommander:
Something is funky here. What is described above is not what the committed code does.

phkb:
Um.....
Yeah. Please look away. I got my code tweaks mixed up. I'll do it again, paying a bit more attention this time!

PR #462, opened 2023-11-15
Hostile lollipops in colorblind mode

AnotherCommander:
This is an accessibility oriented pull request. The idea is to make hostile ships on the scanner more easily distinguishable for colorblind people.

The colorblind mode algorithms we have enhance contrast of displayed colors to make them easily distinguishable as different. Still, sometimes when there is a lot going on in the screen, some additional visual clues would be good for persons who have trouble distinguishing key colors, even if we modify the contrast to make it easier for them. With this PR, in addition to the colorblind filters in use, we change the lollipop blips of hostile ships to an X instead of a square. This way, it only takes one quick look on the scanner and all hostiles can be easily identified.

This is something I had wanted to add for a long time now and is also consistent with color blindness accessibility best practices (don't just change the color, change the shape of key game elements as well if you can).

### DIFF
--- a/Doc/CHANGELOG.TXT
+++ b/Doc/CHANGELOG.TXT
@@ -13,6 +13,9 @@ General:
   screen to be displayed.
 * Added new visual effect when cloaked.
 * Added colourblind support, available in game options screen.
+* In colorblind mode, scanner blips of hostile ships are now drawn as X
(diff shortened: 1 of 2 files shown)


Zireael07:
Not colorblind, but I like the idea of having hostile lollipops have a different shape <3

PR #466, opened 2023-11-24
allowing snapshot on long range chart

phkb:
Not much more to say, really!

### DIFF
--- a/src/Core/Entities/PlayerEntityControls.m
+++ b/src/Core/Entities/PlayerEntityControls.m
@@ -991,7 +991,7 @@ static NSTimeInterval	time_last_frame;
 		// snapshot
 		const BOOL *joyButtonState = [[OOJoystickManager sharedStickHandler] getAllButtonStates];
 		if (([self checkKeyPress:n_key_snapshot] || joyButtonState[BUTTON_SNAPSHOT]) &&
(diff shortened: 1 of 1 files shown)


src/Core/Entities/PlayerEntityControls.m
```diff
@@ -991,7 +991,7 @@ - (void) pollApplicationControls
 		// snapshot
 		const BOOL *joyButtonState = [[OOJoystickManager sharedStickHandler] getAllButtonStates];
 		if (([self checkKeyPress:n_key_snapshot] || joyButtonState[BUTTON_SNAPSHOT]) &&
-			![gameView allowingStringInput] && // not while entering text on the keyboard config screens
+			(![gameView allowingStringInput] || gui_screen == GUI_SCREEN_LONG_RANGE_CHART) && // not while entering text on the keyboard config screens
```
AnotherCommander, line 994: Proposal: Ensure that allowingStringInput is either alphanumeric or nothing, if so then take snapshot. This has the advantage of working also with future screens that may require text input and does not touch keyboard config or load/save screens.
[gameView allowingStringInput] <= gvStringInputAlpha

phkb: Nice! Much better!

PR #468, opened 2023-12-08
Hides the price and install time values on the F3 Ship Outfitting screen

phkb:
This makes it possible to have more display options on the F3 screen (allowing header entries, or top-level entries).


AnotherCommander:
Can't test functionality right now, but code looks good to merge. Feel free if you are happy that it works correctly.

PR #469, opened 2023-12-11
Added player ship property to allow status screen equipment to have a…

phkb:
…scending sort
The default sort method works to put the most important equipment on the first page of the F5 Status screen. The assumption is, if it's got a high TL and price, it must be more important. The sort_order property of equipment is considered before these two. However, because the list is in reverse order, any ordering that has been encoded into the "sort_order" property will also be reversed.
To work around this, a new r/w property has been added to the player.ship: "statusScreenEquipmentSort". The default value for this is 0, which is the reverse order method, as it currently works. By changing this value to 1, the list goes back to an ascending order, which gives OXP's that make heavy use of the "sort_order" property the opportunity to decide what order should be used on the status screen.

### DIFF
--- a/Doc/CHANGELOG.TXT
+++ b/Doc/CHANGELOG.TXT
@@ -39,6 +39,8 @@ General:
 * Improved sun rendering.
 * Implemented user-defined planet dark-side illumination color, when an
   illumination map is in use.
+* Added "statusScreenEquipmentSortMode" to switch the ordering of equipment
+  between the default reverse order and ascending order.
 * Added an "update all" option to the Expansion Pack Manager.
 * Updated atmosphere density rendering.
 * Added game paused message to game options when in flight.
@@ -76,6 +78,8 @@ Scripting:
 * Added planet.airDensity property.
 * Added Ship.setShipInfoForKey to allow for shipdata updates during the game.
 * Added equipment.weaponInfo dictionary.
+* Added player.ship.statusScreenEquipmentSortMode read/write property. 0 = default,
+  (reverse), 1 = ascending.
 
 Bug Fixes:
 ==========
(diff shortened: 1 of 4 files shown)


cim--:
Would it be simpler - and yes, I should have done this myself most of a decade ago! - just to document the existing behaviour of sort_order at https://wiki.alioth.net/index.php/Equipment.plist
"The higher the sort_order number, the earlier on the list the equipment will appear. The default value is 1000. Items with identical sort_order will be sorted by tech level and price (higher appearing first)"

Having to document "either a higher or lower sort_order may make equipment appear earlier on the list, depending on the value of a playership property which any number of OXPs may be competing to set" seems less helpful.

phkb:
> Would it be simpler - and yes, I should have done this myself most of a decade ago! - just to document the existing behaviour of sort_order at https://wiki.alioth.net/index.php/Equipment.plist "The higher the sort_order number, the earlier on the list the equipment will appear. The default value is 1000. Items with identical sort_order will be sorted by tech level and price (higher appearing first)"
> 
> Having to document "either a higher or lower sort_order may make equipment appear earlier on the list, depending on the value of a playership property which any number of OXPs may be competing to set" seems less helpful.

However, on the F3 screen, the sort_order has the opposite effect - the lower the sort_order, the earlier in the list it appears.

cim--:
True, that would need to be included in the documentation - though that also makes sense in general: as equipment will usually disappear from the F3 list once purchased, having the cheaper and lower-tech items which a player will normally buy first at the top is desirable.

If there's a need for certain pieces of OXP equipment to appear high up on both lists then a separate purchase_sort_order property (defaults to the same as sort_order) and changing the core Fuel item to use that might be more reliable.

phkb:
... and a new equipmentEnumerator.

cim--:
Yes, probably - though for how little F3 gets used and how many equipment items are likely, shallow copy to a local variable and re-sort within the F3 display function is probably perfectly fine for efficiency.

PR #472, opened 2023-12-17
Adds consoleMessageReceived event

phkb:
And updates some changelog entries.


AnotherCommander:
Are the changelog changes OK in this PR? Please double check, looks like it is removing mentions of features that went in earlier.

phkb:
Thanks, @AnotherCommander . Should be fixed now.

PR #476, opened 2024-02-26
Added ability to set default activate+mode keys for equipment items

phkb:
With this PR, you can now set "default_activate_key" and "default_mode_key" in your equipment.plist definition. If the chosen key conflicts with another key when loading, the setting will be ignored.
The definition is the same as for other key settings, in that it is an array of dictionary items. For example:
```
	default_activate_key = ({key = "0"; shift = true;});
	default_mode_key = ({key = "9"; mod1 = true;});

```
If you change the default key for the equipment item, the "Overrides" column will now show that the key setting has changed from the default.

### DIFF
--- a/Doc/CHANGELOG.TXT
+++ b/Doc/CHANGELOG.TXT
@@ -65,7 +65,10 @@ Modified Plists:
   install time values on the F3 Ship Outfitting screen.
 * equipment.plist: purchase_sort_order allows the sorting of the F3 Ship Outfitting
   screen to be sorted independently of the F5 Status screen.
-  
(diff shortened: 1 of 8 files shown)


phkb:
I think this is stable and predictable now. The first version would reset the activate/mode keys back to the default even if you'd specifically unset them. Now, the only time it picks up the default keys is when you purchase the equipment for the first time. If you add default keys to an equipment item later, you would need to go into the keyboard settings and reset the keys back to default.

Anyway, I'll push to master in the next day or so unless there's objections.

PR #477, opened 2024-03-14
Fixes issue with too many choices to fit on mission screen

phkb:
This PR fixes the issue where too many choices have been defined in the options dictionary to fit on the current mission screen.

Essentially, what I'm doing here is (apart from putting a notice in the log) ignoring choices that fall outside the scope of the screen. This could potentially leave the screen in a weird state, if (say) the option to close the screen is one of the excluded options.

As an alternative, I could ignore *all* options if too many have been defined. Let me know your preference.

### DIFF
--- a/Doc/CHANGELOG.TXT
+++ b/Doc/CHANGELOG.TXT
@@ -92,6 +92,8 @@ Scripting:
 
 Bug Fixes:
 ==========
+* Prevented native code exception when too many choices have been defined for a 
+  mission screen that will fit on the current screen size.
 * Splinters that have a defined cargo will now spawn with that, instead of 
   defaulting to minerals.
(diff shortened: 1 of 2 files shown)


AnotherCommander:
I think it looks like a reasonable fix. The root cause of the reported thrown exception is almost certainly the NSMakeRange at line 2093 and to stop the exception one must ensure that the parameters to that method are positive integers - which you do with this fix. I feel a bit uncertain about the best way to handle the problem of overflowing options, but if we go with ignore all options, then it would probably be best to write something on the screen along the lines of [options overflow, all ignored] or something like that. This of course has the problem of immediately breaking immersion, since it takes the user out of the game environment by looking like a debug screen. I'll leave the hard choice to you ;-), but I wouldn't object if this PR gets merged the way it is.

PR #482, opened 2024-05-22
Introducing -help / --help Command Line Option

AnotherCommander:
This PR aims to assist new players trying to work their way through the game at its early stages. It enables the -help (--help also works) command line option. Entering [-]-help as a command line parameter opens up a message box with the list of available command line options on Windows, while on Linux it sends all output to stdout. The game will exit if the [-]-help parameter is detected in the command line args, immediately after the help info has been displayed. The exit code in this case is 2, which means exiting due to [-]-help found.

A short message indicating the exit reason is sent to the actual log as well. This is to avoid sudden termination of the log file, which could lead to the potential (wrong) assumption that a crash occurred.

### DIFF
--- a/src/SDL/main.m
+++ b/src/SDL/main.m
@@ -70,7 +70,9 @@ int main(int argc, char *argv[])
 	
 #if OOLITE_WINDOWS
 
-	// Detect current working directory and set up GNUstep environment variables
+	#define OO_SHOW_MSG(ooMsg, ooMsgTitle, ooMsgFlags)	MessageBox(NULL, ooMsg, ooMsgTitle, ooMsgFlags)
+ 	
+ 	// Detect current working directory and set up GNUstep environment variables
 	#define MAX_PATH_LEN 256
 	char currentWorkingDir[MAX_PATH_LEN];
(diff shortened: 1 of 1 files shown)


AnotherCommander:
I would appreciate some feedback from Linux users. Does this work OK for you? Note that the tabs in the command line args output may be a bit off in Linux; I wouldn't worry too much about that, the tabs are calibrated for correct display inside the Windows message box.

src/SDL/main.m
```diff
@@ -129,6 +134,30 @@ int main(int argc, char *argv[])
 				if (i < argc)
 					[controller setPlayerFileToLoad: [NSString stringWithCString: argv[i]]];
 			}
+
+   			if (!strcmp("-help", argv[i]) || !strcmp("--help", argv[i]))
+			{
+				char const *processName = [[[NSProcessInfo processInfo] processName] UTF8String];
+				char s[1024];
+				sprintf(s,	"Usage: %s [options]\n\n"
```
dsalt, line 142: Should be `snprintf` or, ideally, `asprintf`. (Okay, it's safe unless the process name is unusually long, but as a matter of principle…)

PR #494, opened 2024-11-24
New ECM Visual Effect

AnotherCommander:
This PR implements a visual addition to the ECM effect. Till now ECM blast presence is indicated by a jitter of the scanner lollipops. With the changes from this PR we now also get an EM interference pattern on the ship's view screen, expressed as a burst of static and/or CRT malfunction. The effect is available only when shaders are available, so you will be able to experience it in the Shaders Enabled and Extra Detail detail levels. You will get it at every ECM trigger event, be it from NPCs or the player.

The original shader used as inspiration for this is https://www.shadertoy.com/view/ltV3z1 . Note that it seems to rely on some strange code operations like mod(x, 0.0). However, it seems to perform exactly as advertised so I did not make any substantial changes to it.

Linux maintainers, please confirm that it works also for you if you can. I tried to make sure that this effect will be available and functional on Linux too, but need to be sure.

This video shows what the new effect looks like.

[IMAGE ALT TEXT](http://www.youtube.com/watch?v=fQq0Fun1Ty8 "Oolite 1.91 - New ECM Effect")


Garryck:
Nice.. it shuts off very abruptly though.. if the effect could fade out over a half to one second, I suspect it would look even better. Assuming that's even possible, of course.

PR #505, opened 2025-04-19
Added missing normal and specular map entries for rock-hermits

phkb:


AnotherCommander:
Nice catch, good to merge as far as I can see.

PR #521, opened 2025-07-24
Add Clang flag, assume x86_64 for Windows/Linux

mcarans:
Also split GNUmakefile to make it easier to follow

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -13,8 +13,6 @@ GNUSTEP_OBJ_DIR_BASENAME         := $(GNUSTEP_OBJ_DIR_NAME)
 
 ifeq ($(GNUSTEP_HOST_OS),mingw32)
 	vpath %.rc src/SDL/OOResourcesWin
-    oolite_WINDRES_FILES = \
-	    OOResourcesWin.rc
 	
     WIN_DEPS_DIR                 = deps/Windows-deps/x86_64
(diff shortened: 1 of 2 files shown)


mcarans:
Tested on Windows (gcc) and Linux (Clang)

mcarans:
I have removed 32 bit stuff in Makefile and GNUmakefile. I tested building and installing on Linux and Windows. Both worked fine.

AnotherCommander:
Tested it with creating installers for deployment and snapshot versions for Windows and all seems good.

My only comment here would be about the two Windows-only files OOResourcesWin.rc and miniz.c. Any particular reason they were moved out of files.make and into GNUmakefile? To me it would make more sense to have them inside files.make since they are proper project files; right now they are just two stray references in two different places in GNUmakefile when there is a place specifically for storing them. Just add them to files.make right after oolite_CC_FILES with a guard like
```
ifeq ($(GNUSTEP_HOST_OS),mingw32)
    oolite_C_FILES += \
	miniz.c
	
	oolite_WINDRES_FILES = \
	OOResourcesWin.rc
endif
```
and all is good. Unless there is a problem with building Linux, in which case OK, but I think it should work just fine.

mcarans:
@AnotherCommander Thanks for the feedback. I have made the requested changes.

PR #526, opened 2025-08-27
Fixes Linux format-security issues

phkb:
Courtesy of LW-archlinux


AnotherCommander:
Also, this is just gcc being silly to the point of infuriating people. Just had to drop this here.

PR #527, opened 2025-08-28
Adding new world event guiSelectedRowChanged

phkb:
This PR adds a new world event, "guiSelectedRowChanged", which is raised any time the player selects a new item on a gui screen. This could happen by using the up and down arrow keys, or by using the mouse.

The goal of this event is to give scripters the ability to see what the player is doing on a screen and respond to it _before_ the player presses the enter key. For instance, looking towards accessibility development, the text of the selected item could be read out to the player audibly as soon as they select it. 

The format of the event is:

`this.guiSelectedRowChanged = function(key, row, text) {
}`

Where
   key (string) = the underlying key code for the selected row, based on the screen currently shown.
   row (int) = the gui screen row index value for the selected row
   text (string) = the text of the selected row

The event will fire on all screens that have any sort of gui element, including the options screen (F2), the game options screen, and any mission screen. Screens where the event doesn't fire are the various chart screens (F6 and F6F6), the system info screen (F7), the manifest screen (F5F5), but only if there is a single page, and the market info screen (F8F8).

### DIFF
--- a/src/Core/Entities/PlayerEntity.m
+++ b/src/Core/Entities/PlayerEntity.m
@@ -8876,7 +8876,9 @@ static NSString *SliderString(NSInteger amountIn20ths)
 
 	[[UNIVERSE gameView] clearMouse];
 	[[UNIVERSE gameController] setMouseInteractionModeForUIWithMouseInteraction:YES];
-	
+
+	gui_screen = GUI_SCREEN_GAMEOPTIONS;
+
 	// GUI stuff
 	{
 		#define OO_SETACCESSCONDITIONFORROW(condition, row)				\
@@ -9139,7 +9141,6 @@ static NSString *SliderString(NSInteger amountIn20ths)
 	/* ends */
 
 	[self setShowDemoShips:NO];
-	gui_screen = GUI_SCREEN_GAMEOPTIONS;
 
 	[self setShowDemoShips:NO];
 	[UNIVERSE enterGUIViewModeWithMouseInteraction:YES];
@@ -9154,6 +9155,8 @@ static NSString *SliderString(NSInteger amountIn20ths)
 	OOGUIScreenID	oldScreen = gui_screen;
 	
 	[[UNIVERSE gameController] setMouseInteractionModeForUIWithMouseInteraction:YES];
+
+	gui_screen = GUI_SCREEN_OPTIONS;
 
 	if ([self status] == STATUS_DOCKED)
 	{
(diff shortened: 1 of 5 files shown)


AnotherCommander:
Yeah, that's probably a more descriptive name.

phkb:
Also updated the initial description with the new event name and added some additional info about the parameter types.

phkb:
I noticed that when guiSelectedRowChanged fired, sometimes guiScreen wasn't updated to the correct screen. To correct this, I moved the update of guiScreen to be at the beginning of each screen build process, rather than at the end.

AnotherCommander:
This last change bringing the gui screen updates to the top triggers my alarms for some reason. Haven't had more than a cursory glance, but are we sure that the from and to gui screen parameters are not affected by it?

phkb:
I checked every from/to combination I could find, for the ones where I made a change, and couldn't find any instance of where guiScreenChanged reported the change incorrectly. There was some odd code I didn't understand, though: in a couple of instances, this code line was set before and after the guiScreen value was changed:
`[self setShowDemoShips:NO];`
But it's just setting a flag, and guiScreen is just a value; neither of them execute additional code when the value changes, do they?

Or did your alarm bells ring for another aspect of gui screen changes (background, overlays, something else)?

AnotherCommander:
My alarms went off for the gui screen changes initially, but the items you
mentioned are also valid potential fail points that need to be checked.

About the [self setDemoShips:NO];, it might just set a flag, but maybe
further down a method is called that uses the current state of demoShips in
combination with the current gui screen. Haven't looked at the code and I
get possibly paranoid but the point is that just because a flag is changed
without any immediate effects doesn't mean that everything can be
considered safe.

If you are confident that all bases are covered and the risk of moving the
gui screen changes to the top is negligible, we are good to go. Just wanted
to indicate potential points of concern.

phkb:
I've examined the code again, looking at all the function calls between the two points of the gui screens where the change took place: from where I've moved the change to, and to where it was previously, and can't find anything that was dependent on having a specific guiScreen value. I've also tested a few OXP's that depend on having specific values for guiScreen (the "Red Set", Xenon UI, BGS in particular), and they all seem to function correctly. I honestly think I've exhausted all the ways the code *could* break and haven't found any indication that might be happening, so I think we're good. But I'll leave this until tonight in case you have any additional thoughts.

AnotherCommander:
Looks like we are good to go then. Feel free to merge when ready.

PR #532, opened 2025-12-15
Modern linux build with AppImage and source builds for Ubuntu, Fedora and Arch (+ derivatives)

mcarans:


AnotherCommander (commented):
I checked the changes on Windows with the legacy and modern Clang build. All works as expected, no surprises. As far as I can tell the PR is OK for Windows.

Unfortunately I am not qualified to comment on the Linux side of things (which is the largest part of the PR), nor do I have the ability to test it practically. If you have tested it sufficiently and it works for you I guess it's good to go.

mcarans:
@LW-archlinux I have updated the [README.md](https://github.com/mcarans/oolite/blob/modern_linux/README.md).

As for testing the AppImage, I posted for testers [here](https://bb.oolite.space/viewtopic.php?p=303449#p303449). Testers need only download the zip, extract the AppImage file in the zip and run it. Have you tried it on your machine?

I'll see if I can try using Distrobox or Podman to test.

mcarans:
@LW-archlinux I think this is now ready. I've updated the README to mention developers building from source, talk about sudo and reference the bulletin board for people to seek help. 

I think the jsapi isssue is that submodules were not checked out. The README has `git clone --filter=blob:none --recurse-submodules https://github.com/OoliteProject/oolite.git` which will be correct when the PR is merged but you would need: `git clone --filter=blob:none --recurse-submodules -b modern_linux https://github.com/mcarans/oolite.git`. In any case I have added a step in checkout_deps to check if the submodules have been checked out and check them out if not.

Once this is merged, you can add Arch specific instructions including about your Arch package to the README if you like.

@AnotherCommander You probably do this anyway, but when merging, I'd advise squash and merge to avoid lots of noise on master as I had to do a lot of experimentation with GH workflows.

PR #541, opened 2026-01-15
Debug static build on Windows

mcarans:
@AnotherCommander I have tested the installer but not got round to checking the debug build of Oolite although I've checked the necessary libraries are in the Spidermonkey package

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -29,7 +29,7 @@ ifeq ($(GNUSTEP_HOST_OS),mingw32)
     JS_INC_DIR                   = $(WIN_DEPS_DIR)/JS32ECMAv5/include
     JS_LIB_DIR                   = $(WIN_DEPS_DIR)/JS32ECMAv5/lib
     ifeq ($(debug),yes)
-        JS_IMPORT_LIBRARY            = jsdbg
+        JS_IMPORT_LIBRARY            = js32ECMAv5dbg.dll  # to use jsdbg, gcc builds need ADDITIONAL_CFLAGS, ADDITIONAL_OBJCFLAGS: -DSTATIC_JS_API
     else
         JS_IMPORT_LIBRARY            = js
     endif


AnotherCommander:
Gave it a test run with both debug and release builds and here are my observations:
- The debug build fails at the very last step when it tries to copy over the dlls with the ldd command. The problem is that the debug executable is named oolite.dgb.exe rather than oolite.exe, but the ldd command hardcodes it to oolite.exe so it fails on the debug build. Solution: change the ldd command to this: `ldd $(PROGDIR)/$(DEST_BIN) | grep $(MINGW_PREFIX) | awk '{print $$3}' | xargs -I {} $(CP) $(CP_FLAGS) {} $(PROGDIR)` (`$(DEST_BIN)` instead of oolite.exe at the start).

- The MSVCRT build fails at linking. To fix that you need to add the flag `-DSTATIC_JS_API` to additional cflags and objcflags. The Clang linker does not fail and emits a series of warnings instead. The Clang debug executable works fine even with those linker warnings though. To add the STATIC_JS_API flag only to the gcc builds you may want to move the compiler detection at the top of the GNUmakefile so that it is common to both Windows and Linux. This way you could use something like
```
ifeq ($(debug),yes)
    ifeq ($(COMPILER_TYPE),gcc)
        ADDITIONAL_CFLAGS        += -DSTATIC_JS_API
        ADDITIONAL_OBJCFLAGS     += -DSTATIC_JS_API
    endif
 endif
```
- We can avoid all the above hassle if we just link with the dynamic library for the debug build only. Instead of linking against libjsdbg.a, we can link against libjs32ECMAv5.dll.a instead. This avoids any kind of breakage of the MSVCRT build and you don't have to add the static api flag. I would prefer this, because a) it is simpler, b) it is inconsequential, since the debug builds are only done by developers and are not distributed in any shape or form and c) avoids having me insert a 30MB libjsdbg.a to the Windows deps repository to support the debug legacy build. My recommendation would be to just go with 
```
ifeq ($(debug),yes)
    JS_IMPORT_LIBRARY            = js32ECMAv5dbg.dll
else
    JS_IMPORT_LIBRARY            = js
endif
```

mcarans:
Ok I've done 1 and 3.

PR #547, opened 2026-02-01
Updates to headers on various pages to better reflect content

phkb:
Plus changing "primable" to "prime-able" which seems more correct. Happy to change it back if there's a preference, though.


phkb:
Really? This is what I see:
[](https://postimg.cc/0MyJcWbJ)
(LibreOffice 25.8.4.2)

Is that not what you're seeing?

AnotherCommander:
Never mind, I was looking at it on the work computer and that is missing the Deja Sans Condensed font. Sorry, false alarm.

AnotherCommander:
I will double check to confirm on my own PC when I get home. Should be able to merge the PR in a few hours.

AnotherCommander:
Actually no need to wait. I installed the font on my work PC and confirmed that it works fine. PR getting merged right away.

PR #550, opened 2026-02-10
More changes requested by FlatHub

mcarans:
The FlatHub approval is proving to be a bit of a pain. I will need to have a tag once this is merged so I can update it in the Flatpak YAML in the PR on FlatHub. Given that making a new release involves changes in various places and I can't be sure that FlatHub won't come back with more changes, do you think making a GitHub tag such as 1.92.0.1 without a full release is ok? If they come back with more, I might need to make 1.92.0.2 etc.

### DIFF
--- a/src/SDL/MyOpenGLView.m
+++ b/src/SDL/MyOpenGLView.m
@@ -127,16 +127,11 @@ enum PreferredAppMode
 	if (showSplashScreen)
 	{
 #if OOLITE_WINDOWS
-		// Pre setVideoMode adjustments.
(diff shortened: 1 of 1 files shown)


mcarans:
@AnotherCommander As this is on the 1.92-maintenance branch, if I understand correctly, I think that a release will not be created by the workflow: https://github.com/OoliteProject/oolite/blob/1.92-maintenance/.github/workflows/build-all.yaml#L358

I would then create a tag 1.92.0.1 on the version in the branch once you approve/merge the PR. Is that ok?

PR #552, opened 2026-02-24
Version full inline

AnotherCommander:
The full version string is now inlined in the game files and the executable itself.

The macro `OO_VERSION_FULL`, representing the full version string (e.g. "1.93.0.7779-260224-dc76e30"), is now defined at the command line during compile time for all C and Obj-C files in the project. This allows easy access to the version string from practically everywhere.

The full version string is added to the `[-]-help` output, as well as the main menu screen when the `-showversion` command line argument is used. The full version is now also printed in the log header and is now the version reported for all core world scripts.

In addition to all the above, this PR embeds the full version string to the core resources manifest.plist file, replacing the previous `Maj.Min` version format under the `version` key. The build type is also inlined in manifest.plist under the key `debug_functionality_disabled`, which takes the `yes/no` values. `yes` implies a deployment build, `no` implies a test release or a snapshot build.


GNUmakefile.postamble
```diff
@@ -9,11 +9,13 @@ POST_BUILD_ENV = \
...
 	MINGW_PREFIX="$(MINGW_PREFIX)" \
 	DEBUG="$(debug)" \
+	DEPLOYMENT_RELEASE_CONFIGURATION="$(DEPLOYMENT_RELEASE_CONFIGURATION)" \
 	MODERN="$(modern)" \
 	ESPEAK="$(ESPEAK)" \
 	USE_DEPS="$(use_deps)" \
 	STRIP_BIN="$(strip)" \
-	STRIP="$(STRIP)"
+	STRIP="$(STRIP)" \
+	VER_FULL="$(VER_FULL)"
```
mcarans: I can't see any changes to postbuild - is this variable used in there?

Funnily enough, before you made the changes to the manifest, I had planned to pass the DEPLOYMENT_RELEASE_CONFIGURATION variable to the postbuild and use it like this (but now will just output the manifest instead):
```
    if [ "$DEPLOYMENT_RELEASE_CONFIGURATION" = "yes" ]; then
        echo "DEPLOYMENT=yes" >> build/OOLITE_VERSION.txt
    else
        echo "DEPLOYMENT=no" >> build/OOLITE_VERSION.txt
    fi
    cp -fu build/OOLITE_VERSION.txt "$PROGDIR/Resources"

```

mcarans (approved):
Looks great to me. I suggested some small improvements.

PR #557, opened 2026-03-03
Don't assume x86_64 when looking for espeak-ng-data.

dsalt:
On Debian-like systems, the data is stored in an architecture-dependent location, so on aarch64 the usual location is /usr/share/aarch64-linux-gnu.

(Almost certainly safe to assume *-linux-gnu, so this patch still does that.)


mcarans:
I imagine that this will be fine.

mcarans:
Implemented in https://github.com/OoliteProject/oolite/pull/732. Closing.

PR #564, opened 2026-03-21
Changes from 1.92.1.0

mcarans:
Bring over changes from 1.92.1.0. I have tested on Linux and Windows.

I decided it would be better to do this first and make sure it is working ok than to add it to my build reworking which is in progress.


mcarans:
Closing PR in favour of cherry picking.

PR #566, opened 2026-04-01
Remove unused Git submodules, update Linux-deps version

mcarans:
Remove what looks like unused Mac-specific, libogg, libvorvis and tests Git submodules 
Update Linux-deps version
Update Flatpak yaml to reflect FlatHub build
Remove Mac-specific document generation in Doxyfile

### DIFF
--- a/Documentation/Doxyfile
+++ b/Documentation/Doxyfile
@@ -743,7 +743,7 @@ WARN_LOGFILE           =
 # spaces.
 # Note: If this tag is empty the current directory is searched.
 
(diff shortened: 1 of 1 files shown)


AnotherCommander:
I don't have any suggestions either. If in doubt, do not delete the folders in deps, just remove the entries from the two gitmodule files.

AnotherCommander:
Can you please check what happens if you want to checkout a previous git revision before the point of the submodule deletion? Does the repository revert back to exactly what it was before (with the submodule present as it was at the time of the revision)? Can you then checkout HEAD again and move to the latest revision with the submodule reference removed without any issues?

Apart from that point to confirm, looks good to me.

PR #590, opened 2026-04-06
Logic when checking whether station has a launch dock was incorrect

### LINKED ISSUE #589: station.launchPirateShip() never launches ships
The station.launchPirateShip() fails to launch any ship from any station.

To reproduce, start a strict game. A strict game will generate 2 stations in Lave: the main stations, and 1 Rock Hermit. From the Debug Console enter the following command:

> S.stations[0].launchPirateShip()
> S.stations[1].launchPirateShip()

The result for both will be null.

phkb:
Fixes #589


oocube:
With the fact that arch builds currently are failing for another reason this PR looks good to merge.

PR #597, opened 2026-04-11
Windows-deps submodule removed

mcarans:
Windows-deps submodule removed using approach described here: https://bb.oolite.space/viewtopic.php?p=304803&sid=8f1bdc4e93e44a685eda6376b976cf83#p304803

The submodule is not needed for the modern build.


phkb (commented):
Yep, confirmed, the Windows build still works, game plays as normal, and everything still seems to be present. All good from me!

PR #598, opened 2026-04-11
Fast enumeration only and debug JS library

mcarans:
No need for the non fast enumeration with modern build


mcarans:
@oocube This also fixes the debug build on Linux by including the new Linux Spidermonkey debug static library ie.
fixes https://github.com/OoliteProject/oolite/issues/592

oocube:
Looks good to me. It makes the code even smaller and easier to maintain
+1

PR #600, opened 2026-04-11
Fixes issue where GUI_SCREEN_NEWGAME was not being passed to guiScree…

### LINKED ISSUE #587: guiScreenChanged does not distinguish between a new game and a game load
The guiScreenChanged world event does not get the "GUI_SCREEN_NEWGAME" as the "from" parameter. It always reports as "GUI_SCREEN_LOAD"

To reproduce, add the following code to a worldScript:
```
this.guiScreenChanged = function (to, from) {
    log(this.name, "to = " + to + ", from = " + from);
}

```
Start a new game, and the following will be seen in the log file:

> 09:33:59.892 [testingGuiScreenChanged]: to = GUI_SCREEN_STATUS, from = GUI_SCREEN_LOAD

The same is true for every version of Oolite back to 1.82. In 1.80, though, the same code produces this output:

> 09:42:26.899 [testingGuiScreenChanged]: to = GUI_SCREEN_STATUS, from = GUI_SCREEN_NEWGAME

phkb:
…nChanged when starting a new game

Fixes #587


phkb:
Redid this, as I messed up the rebasing on the original PR

PR #603, opened 2026-04-14
Mozilla JS in Linux from GitHub release

mcarans:
Build specific versions of GNUstep (rather than just latest master).

Install GNUstep and Mozilla to either system (/usr/local), home ($HOME/.local) or build folder. Default is $HOME/.local so as not to conflict with any system GNUstep libraries a user might be using for other purposes and to have a single copy of the libraries if the user has multiple oolite builds.

Use local in all Bash functions to prevent environment variable pollution.

This change should allow further deletion of submodules.


oocube:
Although I cannot test/judge, the change makes sense and it builds cleanly and passes the tests.
+1

PR #605, opened 2026-04-19
Build all on pull request and regex fix for /usr/local exclusion

mcarans:
Adds build on PR not just push so we can ensure that for PRs not only that the test runs on all platforms, but that packages can be successfully built
Fixes /usr/local exclusion issue found by LoneWolf https://bb.oolite.space/viewtopic.php?p=305180#p305180


oocube:
This is a very reasonable thing to do.
+1

PR #608, opened 2026-04-21
Update compiler flags and ensure those flags are reflected in built packages: debug symbols for Linux, use lto thin 

### LINKED ISSUE #586: Separate debug symbols file on Linux
Split Dwarf or objcopy extraction result in a "slim" binary and a separate debug file.

For the former, build with -gsplit-dwarf then run dwp to create the master symbol file.

mcarans:
LTO thin gives virtually all the benefit of LTO with fast linking and very fast incremental builds
Added debug symbols for Linux 
Debug symbols in appimage and separate flatpak (as per flatpak spec)
Symbols on Linux are in oolite.debug
On Linux debug build binary is just oolite not oolite.dbg 
Remove unnecessary env var for specifying strip binary (just use strip) and unused flatpak post build file
Copy oolite.debug and Oolite source to Debug flatpak
Fix Arch build

Fixes https://github.com/OoliteProject/oolite/issues/586


mcarans:
Example flatpak debug:

```
> flatpak run --devel --command=gdb space.oolite.Oolite -ex "run" --args sh /app/bin/run_oolite.sh debug
GNU gdb (GDB) 17.1
...
Reading symbols from /app/bin/oolite...
Reading symbols from /usr/lib/debug/app/bin/oolite.debug...
(gdb) break OOOXZManager.m:233
Breakpoint 1 at 0x68fada: file src/Core/OOOXZManager.m, line 233.
(gdb) run
Starting program: /app/bin/oolite -nosplash
...
2026-04-22 09:46:07.397 oolite[36:39] Truncating thread name 'OOLogOutputHandler logging thread' to 15 characters due to platform limitations

Thread 1 "oolite" hit Breakpoint 1, -[OOOXZManager installPath] (self=<optimized out>, _cmd=<optimized out>) at src/Core/OOOXZManager.m:233
233             const char *managedAddOnsEnv = SDL_getenv("OO_MANAGEDADDONSDIR");

```

phkb (commented):
Not much impact from a Windows build perspective, so fine with me. Build still works.

LW-archlinux:
Some questions for clarity:

Which build targets does this affect ?


When compiler has to autodetect debug setting they use different default values (gcc uses 2 as default and clang 0 ) .
These values have  a huge impact on the build resources used and the size of the files with debug symbols.
Have there been tests run to find which value works for olite and set that instead of letting the compiler decide it ?

Why is changing LTO settings added to a PR that deals with debugging symbols ?

Apparently flatpak & appimage require special handling to include debug symbols. Why weren't they dealt with in separate PRs ?

mcarans:
@LW-archlinux It affects all build targets on Linux. For the ones that strip (release and release-deployment), a separate debug symbols file is made. For the ones that don't strip (release-snapshot and debug), the existing symbols in the binary are compressed to drastically reduce file size.

Both GCC and Clang use level 2 as the default when you specify the -g flag without a trailing number. I have now changed to -ggdb3 for targeting the gdb debugger specifically and including macros (on Linux only, Windows still uses -g if it uses it at all). Previously I have found this flag works well for debugging.

I consider this PR a single task to update compiler flags and ensure those flags are reflected in built packages. 

Even if that wasn't the case, it's also a matter of time and drive space. I am a volunteer so have limited time. I'm trying to maximise the value of the time I spend working on Oolite. 

To make multiple tiny PRs and have each reviewed is very time consuming and requires having multiple copies of the Oolite source code checked out each building on the branch for the previous PR on which it would depend which is challenging to manage. I might be prepared to spend my time and energy to do it that way if I was asked to do it and payment was forthcoming for the effort involved.

PR #615, opened 2026-04-29
AnyLinux universal appimage

### LINKED ISSUE #576: Make AnyLinux AppImage instead of traditional one
AnyLinux AppImage - although this has been [prototyped](https://bb.oolite.space/viewtopic.php?p=304680#p304680) using Arch AUR, must be integrated into source builds so that test AppImages can be built on any platform

mcarans:
Fixes https://github.com/OoliteProject/oolite/issues/576

Also fixes broken Fedora test (a missing package).

What's special about the appimage is that it is built on latest Arch so very up to date, but should nevertheless run on older distros because it is using a new type of appimage technology.

### DIFF
--- a/.github/workflows/build-all.yaml
+++ b/.github/workflows/build-all.yaml
@@ -6,7 +6,13 @@ on:
 
 jobs:
   build-linux:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
+    container:
+      image: archlinux:latest
+      options: >-
+        --cap-add SYS_ADMIN
+        --device /dev/fuse
+        --security-opt apparmor:unconfined
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


mcarans:
@oocube Does the appimage work for you?

oocube:
I downloaded https://www.filemail.com/d/qxrazspytredvdq and ran a session. Nothing spectacular there. Sound as ever, FPS as usual and I had voice output as well. Tested on Ubuntu 24 LTS.

There is something I dislike about our builds: I cannot make out which is which. We have such a long version number yet it means quite little to a human being. Especially when builds come from other repositories just like yours only the commit hash in the end would be valid for identification.

mcarans:
Thanks for testing. Yes I agree the build string is overly complicated.

oocube:
I raised #616. I think this PR can be merged.

Issue #578, opened 2026-04-03
Update to SDL3 from SDL1.2


mcarans:

kanthoney:
The `kja_sdl3` branch is my initial attempt at converting to SDL3. It builds and runs in Linux, but I haven't tried Windows. There's a lot of Windows-specific code in there that probably clashes horribly.

Issues I've come across:

- The splash screen doesn't work. If you uncomment the `SDL_UpdateWindow` line in `MyGameView.m`, the splash screen works (except it's squashed horizontally) but the game crashes on startup. 
- The shift key didn't work. It turns out this was due to the scan codes being different in SDL3, so the mappings in `keymappings_linux.plist` and the Windows equivalent were all wrong. I've changed the US and UK mappings in the Linux file but not the other countries or the Windows file.
- There's no `SDL_SetGamma` in SDL3. Apparently we'll have to add a post-processing filter to do this.
- I haven't done anything to the build system other than changing `-lSDL` to `-lSDL3` in `GNUmakefile`.

mcarans:
@kanthoney This is excellent news - well done! I wondered how much of a lift it would be to get SDL3 working - you've managed to do it very quickly.

On the startup screen - this was a real headache for me to launch such that both it and the main Oolite window were centred due to Wayland window handling. What happens now on Linux only is that the start screen is launched by a separate tiny executable [splash_launcher](https://github.com/OoliteProject/oolite/blob/master/ShellScripts/Linux/splash-launcher) from [run_oolite.sh](https://github.com/OoliteProject/oolite/blob/master/ShellScripts/Linux/run_oolite.sh#L59) and the start screen in Oolite itself is always disabled (-nosplash). (The simple SDL2 code for the splash screen is [here](https://github.com/OoliteProject/oolite_splash).)

It would be nice to be able to return to using the built in splash screen with SDL3, but if it continues to be problematic, then we can continue to do it the current way, sidestepping the problem.

As for the Windows specific code, I believe a lot, if not all of it, was added by @AnotherCommander and his input will definitely be needed to understand the changes that were made. Some of the changes may no longer be required for SDL3 and some may have to be done differently - for example HDR is supported directly by SDL3.

kanthoney:
I'm trying to set up a dev environment on my Windows partition, but the setup script is giving me the following error:

```
PS C:\Users\kevin\Documents\git\oolite> .\ShellScripts\Windows\setup.cmd
[...]

PR #620, opened 2026-05-09
Upgrade to SDL3 #578

kanthoney:
Upgrade to SDL3 #578

This pretty much works on my machine, so I figure it's close enough to ready to be tested on a wider range of setups.

I haven't been able to resolve the Linux/Wayland splash screen problem where the game screen ends up largely off screen due to Wayland not allowing programatic window placement, so we'll have to continue with the separate splash screen program work around for now.

On Windows, Oolite starts with the window underneath the command prompt I'm launching from - I'm not sure what's causing that.


phkb:
This is probably me, but I'm getting this error trying to compile locally.

```
src/Core/OOOpenGLOnly.h:58:10: fatal error: 'SDL3/SDL_opengl.h' file not found
   58 | #include <SDL3/SDL_opengl.h>
      |          ^~~~~~~~~~~~~~~~~~~

```

Is there something else I need to do?

kanthoney:
Sorry, you would need to install the SDL3 package, e.g., on Windows, running the `ShellScripts/Windows/install_deps.sh` script would do it

phkb:
OK, install script has been run. The package now builds successfully. But the resultant executable doesn't run. I get a brief switch of the mouse pointer to an hourglass, then back to normal, and nothing appears. No splash, no game window. No error files in the oolite.app folder. No log files created.

kanthoney:
Does it run from the command line within the MSYS2 environment?

```
cd oolite.app
./oolite.exe
```

phkb:
It does! Why would it be different running the app from Explorer?

mcarans:
> It does! Why would it be different running the app from Explorer?

I'm guessing a missing dll. 

This is a busy week for me - I hope to look at this PR in the weekend.

kanthoney:
I wondered whether something still required the old SDL dll. But it's not listed when I run `ldd oolite.exe` and copying it manually into `oolite.app` didn't resolve things, so I guess it's not that.

mcarans:
Compiling on Linux, I got this error:

```
Compiling file src/SDL/MyOpenGLView.m ...
src/SDL/MyOpenGLView.m:1717:6: error: call to undeclared function 'SDL_SavePNG'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
 1717 |         if(!SDL_SavePNG(tmpSurface, [pathToPic UTF8String]))
      |             ^
src/SDL/MyOpenGLView.m:1717:6: note: did you mean 'SDL_SaveBMP'?
/usr/include/SDL3/SDL_surface.h:576:34: note: 'SDL_SaveBMP' declared here
  576 | extern SDL_DECLSPEC bool SDLCALL SDL_SaveBMP(SDL_Surface *surface, const char *file);
      |                                  ^

```

I read that SDL_SavePNG is not in the core SDL3 library so I installed:
```
sudo apt install libsdl3-image-dev

```

However, it still failed. I then tried adding the include and lib:
```
#include <SDL3_image/SDL_image.h>
```

```
 -lSDL3_image
```

However, it still didn't build successfully.

mcarans:
Now got it to build, run, make png snapshot and fly around with:

MyOpenGLView.h:
```
#include <SDL3/SDL.h>
#include <SDL3_image/SDL_image.h>
#define SDL_SavePNG(surface, path) IMG_SavePNG(surface, path)
```

and 

GNUmakefile:
```
    ADDITIONAL_OBJC_LIBS         += -lGLU -lGL -lX11 `pkg-config sdl3-image --libs` -lgnustep-base -l$(LIBJS) -lopenal -lz -lvorbisfile -lpng `nspr-config --libs` -lstdc++
    ADDITIONAL_OBJCFLAGS         += -DLINUX -DXP_UNIX `pkg-config sdl3-image --cflags`
    ADDITIONAL_CFLAGS            += -DLINUX `pkg-config sdl3-image --cflags`
```

Great work @kanthoney !

NB: It's better to rename all the SDL_SavePNG to IMG_SavePNG unless Windows specifically expects the former.

kanthoney:
That's weird. It compiles fine on my Fedora laptop.

The [documentation](https://wiki.libsdl.org/SDL3/SDL_SavePNG) says the `SDL_SavePNG` function is in `SDL3/SDL_surface.h`. What happens if you include that header file instead?

kanthoney:
It turns out the `SDL_SavePNG` function was added in version 3.4, and some distros (e.g. Debian) are still on 3.2

mcarans:
I'll see if I can build 3.4 from source then. Actually probably better I just upgrade to 26.04LTS which has 3.4.

mcarans:
I've made a PR to the kja_sdl3 branch: https://github.com/OoliteProject/oolite/pull/643

sdl-config is a SDL1.2 command so build flags need updating. I've taken the opportunity to clean them up for Linux. A similar fix is needed for Windows (and flag cleanup is desirable too).

ADDITIONAL_CFLAGS are automatically added to C++ flags but not ObjC flags so some flags are superfluous.

kanthoney:
I've merged the pull request, and tidied up the Windows build a little bit. Somebody with more Windows/MSYS2 knowledge than me should have a proper look at it, and also try to find out why it's not running from outside the MSYS2 environment.

kanthoney:
It seems that the reason for the startup crash in Windows when run outside the MSYS2 environment is that the `SDL_getenv` function is case sensitive in SDL3. Under MSYS2, the variable is `PATH` and works fine, whereas for Windows it's `Path`, so Oolite can't find the path in the `main` function and crashes. I guess I'll try one and if that doesn't work try the other.

phkb:
OK, latest patch allows the app to be run generally. I build the app and copied the results over an existing installation that has a full set of OXP's installed (both in AddOns and ManagedAddOns), and an existing OoliteDefaults.plist in the default location. A couple of things I noticed:
1. The default log file location has changed from "C:\Oolite\oolite.app\Logs" to "C:\Users\\{username}\Logs"
2. The default location for ManagedAddons has changed from "C:\Oolite\oolite.app\GNUstep\Library\ApplicationSupport\Oolite\ManagedAddOns" to "C:\Users\\{username}\GNUstep\Library\ApplicationSupport\Oolite\ManagedAddOns", even when the original folder was present and populated with OXZ files.
3. The default location for "OoliteDefaults.plist" has changed from "C:\Oolite\oolite.app\GNUstep\Defaults" to "C:\Users\\{username}\GNUstep\Defaults", even when an existing file was present in the original location.

I'm assuming these defaults can be tweaked in OoliteDefaults.plist, but if the defaults file itself in now centralised, how can we have different setups for different installations?

kanthoney:
@phkb the problem with default locations should be fixed with the latest push

oocube:
> I'm assuming these defaults can be tweaked in OoliteDefaults.plist, but if the defaults file itself in now centralised, how can we have different setups for different installations?

Take a look at https://bb.oolite.space/viewtopic.php?p=305556#p305556, which redirects to https://oolite.readthedocs.io/en/latest/. There are environment variables documented - I am just not sure what OS they apply to.

phkb:
Not sure what's going on, but...
[YouTube Flickering textures in 1.93 sdl3 build](https://youtu.be/IFt9VHXSpYg)

In the linked YouTube video, I've used the [Test Ship Model oxp](https://wiki.alioth.net/index.php/Test_Ship_Model) and selected one of the [Generation Ships](https://wiki.alioth.net/index.php/Generation_Ships_OXP), generation-ship-1 specifically. I haven't zoomed in or out or moved the model in any way yet. I have just opened the Test Ship Model screen, and entered generation-ship-1. For some reason the texture along the frame is flickering. Zooming in and out alternatively stops and restarts the flickering.

Generation ships are using the default shaders, with diffuse, normal, specular and illumination maps. I'll experiment with various combinations of these to see if any particular map is the culprit. This is also with my (rather large) suite of OXP's installed. I've run a test with just the Generation ships and the Test Ship Model OXP's and the flickering is the same.

kanthoney:
Looks like the flickering is also in master, and was introduced when #153 was merged. I'll investigate but it's nothing to do with this PR

kanthoney:
@phkb I've opened a separate issue for the flickering issue (#646)

mcarans:
FYI I've spent some time trying to get the test to work with an Ubuntu 26.04 container (long list of failed attempts here: https://github.com/mcarans/oolite/actions). GitHub is currently still on 24.04 so like Fedora and Arch, a container must be used, but something is different about it compared to running locally on Kubuntu 26.04 where the test works.

mcarans:
PR to fix Ubuntu test and flatpak build: https://github.com/OoliteProject/oolite/pull/653

oocube (dismissed):

mcarans:
A temporary error. The builds all succeed now with a change that was simply to kickstart the build. Windows uses the latest 0.1.3 dependencies created today that use latest GNUstep libraries (which include the PR https://github.com/gnustep/libs-base/pull/614 I made to fix compilation errors that occur if you build GNUstep Base without -Wno-incompatible-pointer-types on Windows).

mcarans (dismissed):
I have done very basic testing (ie. install start choices OXP, check I can see the new choices, launch ship) of appimage (including joystick), Flatpak and Windows build (including HDR).  All were fine.

oocube (dismissed):

phkb:
I noticed that on my joystick (a basic Logitech F310 gamepad), the right stick axis appear to be swapped. that is, up/down used to be "Stick 1 axis 4" and left/right used to be "Stick 1 axis 5", but now up/down is "Stick 1 axis 5" and left/right is "Stick 1 axis 4". Is this expected behaviour?

mcarans:
@phkb I asked AI and it said:

Yes, this is **completely expected behavior** when moving from SDL 1.2 to SDL 3. 

While it feels like a bug or a random driver regression, it is actually the result of a massive, intentional overhaul in how SDL handles game controllers that happened between SDL 1.2 and SDL 2.0, and has been further refined in SDL 3.

---

### 1. The Death of Direct OS Axis Mapping
In **SDL 1.2**, the library did very little processing on joystick axes. It essentially asked the underlying operating system (like the Linux joystick subsystem or Windows DirectInput) for a list of axes and handed them to you in the exact order the OS reported them.

Because every controller manufacturer wired up their hardware differently, "Axis 4" might be the right stick Y-axis on a Logitech F310, but on a PlayStation or Xbox controller, it could be a trigger or the right stick X-axis. 

### 2. The Introduction of the Gamepad API
Starting in SDL 2 (and modernized in SDL 3), SDL introduced the **Gamepad API** (formerly `SDL_GameController`). Instead of forcing developers to manually guess which axis belongs to which stick, SDL built an internal database of hardware mappings. 

To ensure consistency across *all* different types of controllers, SDL established a standardized internal layout based on the standard Xbox/XInput layout:

* **Left Stick X** -> `SDL_GAMEPAD_AXIS_LEFTX` (0)
* **Left Stick Y** -> `SDL_GAMEPAD_AXIS_LEFTY` (1)
* **Right Stick X** -> `SDL_GAMEPAD_AXIS_RIGHTX` (2)
* **Right Stick Y** -> `SDL_GAMEPAD_AXIS_RIGHTY` (3)

Because SDL now enforces this uniform layout across Windows, macOS, and Linux, it re-maps the raw OS joystick indexes to fit this predictable structure. 

For the Logitech F310 (especially if it is switched to **DirectInput/D-mode** on the back), the raw OS axis indices for the right stick are inherently flipped compared to standard XInput. SDL 3 actively corrects this so that **X always comes before Y** in the gamepad mappings, which is why your axis indices swapped places.

---

### How to Handle This in SDL 3

* **Use the Gamepad API (Recommended):** If your game uses standard dual-stick controllers like the F310, you should stop using the raw `SDL_Joystick` functions and switch to the **`SDL_Gamepad`** API. If you open the gamepad and query semantic constants like `SDL_GAMEPAD_AXIS_RIGHTX`, SDL handles the hardware-specific mapping automatically.
* **Stick with Raw Joysticks:** If you are building a simulation game that *must* use raw, unmapped joystick inputs (like flight sticks or steering wheels), you can continue to use `SDL_Joystick`. However, you will need to implement an in-game input configuration menu so players can manually bind their axes, because SDL 3's underlying driver hints will still cause raw axis indexing to differ from the 15-year-old SDL 1.2 behavior.

Does that fit what you found?

phkb:
Yeah pretty much! Glad to know it isn't just me, and worth adding a note somewhere so players know what's going on. Probably in the release notes is the best spot.

phkb:
It looks like Gamma correction has been removed entirely? Is there no equivalent in SDL3?

mcarans:
@phkb See @AnotherCommander's  comment: https://github.com/OoliteProject/oolite/issues/578#issuecomment-4337698687
"The gamma option we have in the game options should probably be retired completely. It was added back when we had no idea about proper gamma correction and it more or less did its thing, but now I don't really think we need it anymore and I've seen it does not work on multi monitor setups so its usefulness in effectively non-existent today. If we do want brighter images we can use exposure or light source radiance in the shaders to achieve even better results."

phkb:
Ah, got it. Thanks!

phkb (dismissed):
I've tested SDR/HDR, joystick functions, keyboard functions, non-US keyboards. It all looks good to me.

oocube:
Testing this version (which is so much easier now that we have them all as prereleases) I find the display be off in a strange way. My feeing is the screen is distorted, and the effect is that my view is slightly to the left instead of where it should be. Here are some screenshots.

Screenshot from 2026-05-25 07-40-05

Screenshot from 2026-05-25 07-40-25

Screenshot from 2026-05-25 07-41-16

Screenshot from 2026-05-25 07-41-31

oocube:
Strange. While the first three startups of the game were distorted as above (it took me that long to make the screenshots), I ran it again to see whether in combat you also need to aim elsewhere. And all of a sudden the complete behaviour is normal again. Not sure where the effect came from but I cannot reproduce it.

mcarans:
Guess we're good to merge then?

mcarans:
@kanthoney Please go ahead with merge if you're ready

kanthoney:
@phkb Can you check something for me with the non-US keyboards? I think SDL3 just knows what keyboard layout you're using so I don't think the keyboard layout setting in game options is doing anything any more. If it's not doing anything I might as well remove that option

oocube:
I have an international keyboard with german layout. It looks pretty much like this one:
https://en.wikipedia.org/wiki/German_keyboard_layout#General_information

With oolite_dev-1.93.1-PullRequest620.174-x86_64.AppImage I went to the keyboard settings and assigned the 'y' key to a function. It was properly acknowledged as 'y' on the screen. Then I assigned '#' to the same function. Again that was properly acknowledged as '#' on the screen. In the game I could use the function with that key.

Any specific test that you are after?

kanthoney:
What I want to know is, if you change this setting to, say, Greek, does it have any effect or does it still think the keyboard is German? I think it'll still think the keyboarrd is German, in which case I can safely remove the setting altogether

image

oocube:
Oh, I overlooked the first line. My german keyboard first of all is detected as Default (US).

grafik

Then I switched to german keyboard and checked for some function on e.g. # and Shift+Z.

Screenshot from 2026-05-26 22-54-00

But although the configuration shows to use # and Shift-Z to zoom and unzoom, I had to use # and Shift-Y.
So something is still fishy.

oocube:
Comparison to oolite_dev-1.93.1-70-x86_64.AppImage:

grafik

I can zoom correctly using the indicated keys. So this PR introduces the fishy behaviour. Do not take any offense. The PR is massive, and almost everything works. :-)

mcarans:
@oocube What happens if you change your keyboard setting in the OS - does Oolite use the correct keys? 

If I understand correctly, the behaviour change you see is expected: with SDL3, the "Current keyboard layout" setting in the game is no longer needed (and should be removed) because Oolite instead uses the OS setting.

mcarans:
@kanthoney I have tried switching to the English Dvorak keyboard in the OS and launching the game. 

The key P on a Dvorak keyboard is where R is on a Qwerty keyboard. In Oolite, I had to press R (Dvorak P) to pause.

Accelerating is W in the Oolite key configuration. Pressing Dvorak W (which is on the bottom row of keyboard) accelerated (not Qwerty W). 

Presumably this is what you expect?

phkb:
The current keyboard setting changes which section of the keyconfig2.plist Oolite will use. Keyconfig2.plist has a bunch of tweaks to various keys so that each keyboard type matches the OoliteRS.pdf file, and so there is consistency in player experience, without having to manually tweak the keyboard layout themselves. It doesn't have anything to do with SDL or what keyboard is set in the OS.

phkb:
I won't have a chance to explore this in more detail until tomorrow. I'll report back then

mcarans:
> The current keyboard setting changes which section of the keyconfig2.plist Oolite will use. Keyconfig2.plist has a bunch of tweaks to various keys so that each keyboard type matches the OoliteRS.pdf file, and so there is consistency in player experience, without having to manually tweak the keyboard layout themselves. It doesn't have anything to do with SDL or what keyboard is set in the OS.

By this do you mean that if I play on a Dvorak keyboard, the key positions should be the same as on a Qwerty keyboard eg. accelerate which is W on a Qwerty keyboard is instead mapped automatically by Oolite to , (comma) on a Dvorak keyboard because comma on Dvorak is in the same position as W on Qwerty (ie. below 2 and 3)?

kanthoney:
@mcarans I'm expecting P to pause on any keyboard layout, regardless of where the P key is located, so that sounds right to me

oocube:
When I hacked my first PC with MS DOS and assembly language, there were two ways to access keyboard input:
1. Go for the make and break codes. Every key, when pressed emits a make code. When released it sends a break code. When held for long it would generate multiple make codes. With this pattern software can detect multiple keys being pressed simultaneously although there were only specific groups where this would work. Make and break codes would differ by little as they both would be based on the key's scan code. The disadvantage was that these keys could be mounted in different locations on the keyboard depending on the keyboard hardware's layout.
2. Go for the characters emitted by the keyboard driver. That means the keyboard driver would react to the make/break codes and generate the characters that were typed. It takes away the wondering where a key is on which layout - that is in the keyboard driver configuration. Also some national specific characters require a sequence of keys to be pressed to get one character. Just look at these french accentuated characters, or japanese keyboard input. 

With the keyboard driver in the loop you could now distinguish raw (make and break codes) and cooked modes.

Why am I mentioning all this? Even today, when looking at libraries that abstract away the hardware there are traces of this behaviour: As an example, mind the keyPressed, keyReleased (both raw) and keyTyped (cooked) methods in [Java AWT](https://docs.oracle.com/en/java/javase/21/docs/api/java.desktop/java/awt/event/KeyListener.html). And I suspect SDL is no different here.

That means we should not only look at what the user does but also how the application processes these events. How is that implemented in Oolite?

Edit: https://wiki.libsdl.org/SDL3/BestKeyboardPractices
This is a worthy read to have a quick overview.

kanthoney:
@oocube is the keyboard layout in your OS set to German? If I set my keyboard layout to German in my OS everything works as I'd expect, i.e. to zoom I have to press where the Z key would be on a German keyboard

oocube:
> @oocube What happens if you change your keyboard setting in the OS - does Oolite use the correct keys?
> 
> If I understand correctly, the behaviour change you see is expected: with SDL3, the "Current keyboard layout" setting in the game is no longer needed (and should be removed) because Oolite instead uses the OS setting.



> @oocube is the keyboard layout in your OS set to German? If I set my keyboard layout to German in my OS everything works as I'd expect, i.e. to zoom I have to press where the Z key would be on a German keyboard


Answering both:
i see little use in twiddling OS wide settings back and forth just to start one or the other application. I know today's OSes allow changing the driver per window - I use that at work on a US keyboard when exceptionally writing german text and need the special characters. Back to home...

My Ubuntu is configured for german keyboard, and I do not switch the layout just to write english. Thus, Oolite always perceives the german input driver settings. In my above experiments I just switched the keyboard inside Oolite. Note my comment before I switched: At that time Oolite was displaying "Default (US)", and the keys behaved well. To me it looked like Oolite tried to apply some mapping for the keyboard which SDL may have covered already.

oocube:
Oh rats. Now that I started fooling around the keyboard combinations, I no longer can request the docking clearance or perform the fast docking. Shift+L and Shift+C are no longer effective despite they are mentioned in the UI and I tried both within Oolite on Default (US) and German. What is going on here?

After all the autoilot activates when pressing "c".

kanthoney:
@oocube Can you try the latest commit?

oocube:
I tested based on oolite_dev-1.93.1-PullRequest620.175-x86_64.AppImage.
Now I was able to go back into the station using Shift+L and Shift+C.
Also I verified this now works regardless whether I am on German or Default (US).

Well done!

oocube (approved):

mcarans:
@kanthoney I noticed that on Windows speech stopped working for me since the SDL3 upgrade. Please can you confirm if it works for you. Linux is fine.

mcarans:
I've debugged the Windows sound issue - requires fixing pcaudiolib due to SDL3 change:

```
struct audio_object *
create_xaudio2_object(const char *device,
                      const char *application_name,
                      const char *description)
{
	HRESULT hr = CoInitializeEx(NULL, COINIT_MULTITHREADED);
	if (FAILED(hr) && hr != (HRESULT)0x80010106) {
		return NULL;
	}
...
```

PR #637, opened 2026-05-16
skip releases for PRs from forks

oocube:


mcarans:
@oocube I suggest adding anothercommander

PR #644, opened 2026-05-17
prevent parallel and duplicate executions to reduce CPU minutes

oocube:
Github fires different events, and some of them so fast that we get duplicate executions of the same workflow on the same commit. With this change only one such invocation would run through while duplicates get cancelled.


mcarans:
Good idea!

Issue #631, opened 2026-05-14
Security implications of a public repository


oocube:
Oolite (and OoliteProject in general) hosts a public repository. Anyone can clone it, perform changes and bring them back in form of Pull Requests. This per se allows the public to contribute to the project.

This on the other hand includes a security risk. There are multiple attack vectors (which I do not want to list right here at this moment).
What is our posture to such risk? What countermeasures are we happy with?

LW-archlinux:
Set up a structure that ensures multiple people have to agree on changes .

A possible implementation using a 3 tier setup

Dev roles : Junior , Senior, Veteran
PR classification : minor, major, critical

Upgrading to a higher classification should be doable for all dev tiers.
Downgrading major to minor would require senior or higher. Downgrading from critical would require veterans.

Merging would require multiple reviews, a minor change could mandate review by 1 juniors + 1 senior/vet .
major change could require 1 senior + 1 vet.

oocube:
This would work to decide whether we want to merge or not. By that time something has already happened...

We need to understand each and every code line. We would need to ensure Oolite still runs with that change. Essentially we would build and run the project. I'd like to have the build and test-run to be automated, even before a human being looks at it. But that is the point: When executing the test, this new, untrusted code already executes. Even more, a PR can change our workflows and run havoc there, in our beloved project.

So how can we run a workflow and build/test the code with minimal security risk?

oocube:
Here are two resources that even reference more regarding the topic:
- https://blog.gitguardian.com/github-actions-security-cheat-sheet/
- https://github.blog/security/supply-chain-security/four-tips-to-keep-your-github-actions-workflows-secure/

My worst concern to all the patterns of what you should do or not in the workflow is that the workflow files are part of the repository and can be changed by anyone using a PR. Yet they get executed in our project scope and using our secrets. I know that is possible since I created, modified and tested the workflow in a side branch, then ran PRs before finally merging onto the main branch.

Google Gemini, when asked "In a public Github repository, how can I prevent others to tamper the workflow files in pull requests? " responded this:

> To protect your GitHub Actions workflow files from being tampered with or maliciously executed via pull requests in a public repository, you need a multi-layered approach. Because anyone can fork a public repo and submit a PR, you must prevent both the merging of tampered files and the unauthorized execution of those files.
> 
> Here is the best-practice setup to secure your workflows.
> 1. Require Approval to Run Workflows (Crucial for Public Repos)
> 
[...]

PR #648, opened 2026-05-20
Add catchall (improve #631)

oocube:


.github/CODEOWNERS
```diff
@@ -1,4 +1,10 @@
 # Require specific reviewers for any changes to GitHub Actions workflows
 # https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
 
+# 1. Default owners for everything
+* @oocube @mcarans @phkb @mwerle @Kaks @AnotherCommander @ericwalch @dsalt @cim-- @KonstantinosSykas @kanthoney @jobi-wan
+
+# 2. Secure our workflows
 /.github/ @oocube @mcarans @phkb
```
mcarans, line 8: Presumably many of the developers listed under default owners are long time contributors to Oolite? If so, why not remove the separate shorter list of developers for the folders? Then any change anywhere including workflows can be approved by any one of the devs who feels able to look at it.

mcarans (changes requested):
Presumably many of the developers listed under default owners are long time contributors to Oolite? If so, why not remove the separate shorter lists of developers for the folders? Then any change anywhere including workflows can be approved by any one of the devs who feels able to look at it.

.github/CODEOWNERS
oocube: My idea of securing the workflows is to allow approvers who understand what they are approving. Not all developers have been in touch with the CI/CD functionality.

On the other hand stuff that does not touch the workflow automation should not be restricted - hence I placed all those who have write access.

phkb (approved):
I'm in two minds with the requested change from @mcarans. Happy to go either way. It probably doesn't need resolution immediately, and given the current threat to CI/CD pipelines, tighter security is the better outcome.

mcarans (approved):

mcarans:
Ok, let's go with it as is.

PR #653, opened 2026-05-23
Fixes SDL3 Ubuntu test and Flatpak build

mcarans:
Ubuntu 26.04 has not been released as a GitHub runner, so using a container to get it.
 
SDL3 has an offscreen driver mode which means that using xwfb-run and all the Wayland software driver stuff isn't needed, a nice improvement.

Flatpak requires building SDL3 to get 3.4 (3.2 is the default).

The Flatpak YAML version stuff is more robust as it is now linked to a comment rather than a specific line number


phkb (approved):
This is fine with me, but all I can do is validate the Windows build works correctly.

kanthoney (approved):

PR #657, opened 2026-05-27
Restore original keypress logging

kanthoney:
Fix the keypress logging from the SDL3 merge (remove logging that shouldn't be there and restore some old logging)


mcarans (approved):
Looks fine to me

PR #672, opened 2026-06-09
Fix GCC flags and refactor meson

mcarans:
gcc build is run like this: `make release-snapshot NATIVE_FILE=gcc.ini`

gcc build tested on Arch VM - see https://bb.oolite.space/viewtopic.php?t=22137


oocube (approved):
I tested oolite_dev-1.93.1-PullRequest672.185-x86_64.AppImage and it is looking good.

Issue #677, opened 2026-06-11
linux packageinfo leaks local sourcedir


LW-archlinux:
oolite trunk rev 5083ea4 , sourcebuild using aur packaga oolite-git

```
$ oolite-git.sh packageinfo
{
    title = "Oolite core";
    identifier = "org.oolite.oolite";
<snip>    
    git_remote_url = "/home/panoramix/Downloads/sourcefiles/oolite-git";
<snip>
}
$ 
```

/home/panoramix/Downloads/sourcefiles/oolite-git is a local cache for downloads used by archlinux makepkg tool .
It should NOT be listed anywhere in oolite info or executable .

I understand oolite can't handle every used packaging tool, a method to set the git_remote_url at build time would suffice to solve this.

mcarans:
@LW-archlinux I am not sure how to handle this. I want a way to distinguish between builds that were done with a clone of the original Oolite repo and builds done from forks as i don't want to end up wondering why I can't find a commit hash for a build only to find that it wasn't created using the main repo. This is what I have right now:
```
echo "    git_remote_url = \"$(git config --get remote.origin.url)\";"
echo "    git_commit_hash = \"$(git rev-parse HEAD)\";"
```
Do you know of any git command that will return the actual repo used without returning that local cache one?

oocube:
Just guessing:
If the git_remote_url does not qualify as URL, it must be a local path.

If we see that situation, we could use a relative path. We could use the username and hostname and path (to be concise but that would resemble the leak that lw-archlinux is trying to prevent). We could use nothing as well. After all a local build means the files could be tampered, so any 'real' version number might be misleading.

That's where I am quite content to simply stay on the old numbering scheme.

mcarans:
@LW-archlinux AI suggested the following:

```
# 1. Get the immediate remote URL (could be a web URL or a local path)
RAW_URL=$(git config --get remote.origin.url)

# 2. Check if the URL is actually a local directory path
if [ -d "$RAW_URL" ]; then
    # If it's a directory, query its internal .git config for the original remote
    # We use || true to prevent the script from crashing if the cache folder is missing/inaccessible
    UPSTREAM_URL=$(git --git-dir="$RAW_URL/.git" config --get remote.origin.url 2>/dev/null)
    
    # If we successfully found an upstream URL inside the cache, use it. 
    # Otherwise, keep the local path as a fallback.
    if [ -not -z "$UPSTREAM_URL" ]; then
[...]

PR #680, opened 2026-06-17
Fix system install of package and use ../share/oolite folder for Resources and AddOns

mcarans:
Fix system installation of Oolite with meson - separate assembling folder after compilation (post build) from installation which now has a separate install script
Use `../share/oolite` folder for Resources and AddOns for system install, Arch packages, appimage and flatpak (will also help with anyone wanting to make Debian or Fedora packages in future)
Fixes https://github.com/OoliteProject/oolite/issues/679

Lower case local variables in scripts
Fixes https://github.com/OoliteProject/oolite/issues/678

Note that like Windows, Linux now changes directory to where the executable resides in the ObjC

Main things to test are debug console works, folders are found correctly eg. for addons etc. (worked for me on Linux)

If you've built snapshot, you can do something like the line below to install:
```
meson install -C build/meson_snapshot --destdir /tmp/test
```

That would install oolite and run_oolite.sh to `/tmp/test/usr/local/bin`, and Resources and AddOns to `/tmp/test/usr/local/share/oolite` (since the default prefix, bindir and datadir of meson are used - these can be overridden).

PR #681, opened 2026-06-18
Local semver

### LINKED ISSUE #676: linux packageinfo version is inconsistent with git tags
oolite trunk rev 5083ea4 , sourcebuild using aur packaga

```
$ oolite-git.sh packageinfo
{
    title = "Oolite core";
    identifier = "org.oolite.oolite";
    
    version = "1.93.0.7945-260611-5083ea4";
<snip>
}
$ 
```

Looking at src/oolite-git/ShellScripts/common/get_version.sh this could be due to not setting $SEMVER or $PROJECTNAME at build time.

oocube:
Fixes #676 by introducing semver to local builds and adding the BUILDER environment variable.


ShellScripts/Linux/download_github.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
```
mcarans: It's great you made a helper script for this. It could probably be put in `ShellScripts/common` as it looks like it could work on Windows MSYS2 as well. 

This should be made a bash function to avoid spawning a subshell. An example is this existing function: https://github.com/OoliteProject/oolite/blob/master/ShellScripts/Linux/dep_location_fn.sh#L3 which is then called like this: https://github.com/OoliteProject/oolite/blob/master/ShellScripts/Linux/build_gnustep.sh#L11-L12. 

I add _fn to the script name ie. download_github_fn.sh to distinguish that it is a function to be called by other scripts.

If you want to also be able to call it like a script for your own use in bash, then you can make a separate download_github.sh that sources the function from download_github_fn.sh (but the install_packages_root.sh script should use the function directly to avoid making a subshell).

.github/workflows/build-all.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
```
mcarans: With gitversion called in the build itself in get_version.sh, why is the dependency on common still needed? We need to get rid of the dependency on common so that each Windows and Linux build is self contained otherwise it's hard to replicate build failures that occur in CI locally. If it is informational, then it can run without being a dependency (remove the  `needs: [common]` lines) in which case it should be renamed to something like ci_information.

src/SDL/MyOpenGLView.m
```diff
@@ -145,9 +145,9 @@ - (NSMutableDictionary *) getNativeSize
 - (NSString*) getWindowCaption
 {
 #ifdef BUILD_DATE
-	NSString *caption = [NSString stringWithFormat:@"Oolite v%@ - %s", @OO_VERSION_FULL, BUILD_DATE];
+	NSString *caption = [NSString stringWithFormat:@"Oolite v%@ by %@ - %s", @OO_VERSION_FULL, @OO_BUILDER, BUILD_DATE];
```
mcarans, line 148: If OoliteProject/oolite is forked and a package is built by CI for the fork, what will OO_BUILDER be? It shouldn't be OoliteProject or whatever name is associated with the main repo.

ShellScripts/Linux/install_packages_root.sh
```diff
@@ -14,6 +15,13 @@ run_script() {
     local SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" &> /dev/null && pwd)
     pushd "$SCRIPT_DIR"
 
+    # install gitversion
+    GITVERSION_TGZ=$(./download_github.sh --owner GitTools --repository GitVersion --filter linux-x --outputdir ${SCRIPT_DIR})
```
mcarans: `$()` creates a subshell - `source download_github_fn.sh` and call a function in there called `download_github` instead

mcarans (dismissed):
Thanks for looking into this. It's a step in the right direction. See my comments for some requested changes

ShellScripts/Linux/download_github.sh
oocube: The current implementation gets called like any other tool that we use. What is so bad about subshells?
From what I read the separation into a an extra bash is an advantage as it rules out side effects.

.github/workflows/build-all.yaml
oocube: For sure we need the 'common ' job. You answered that with a blow question.

src/SDL/MyOpenGLView.m
oocube: That is exactly what the common steps finds out.
https://github.com/OoliteProject/oolite/blob/localSemver/.github/workflows/build-all.yaml#L32

If you demand that the whole chain of commands be run the same locally you remove any possibility to identify the different builds. Right now all it does is define a couple of env vars to be passed into the other jobs. It should be a small enough change we can overcome when building locally.

mcarans: This is best achieved by shell commands I outlined in the other issue (https://github.com/OoliteProject/oolite/issues/677#issuecomment-4704840742) for getting the remote url: 
```
# 1. Get the immediate remote URL (could be a web URL or a local path)
RAW_URL=$(git config --get remote.origin.url)

# 2. Check if the URL is actually a local directory path
if [ -d "$RAW_URL" ]; then
    # Dynamically find the correct git directory without subshells
    if [ -d "$RAW_URL/.git" ]; then
        RESOLVED_GIT_DIR="$RAW_URL/.git"
    else
        RESOLVED_GIT_DIR="$RAW_URL"
    fi
    
    # Query the resolved git directory for the original remote
    UPSTREAM_URL=$(git --git-dir="$RESOLVED_GIT_DIR" config --get remote.origin.url 2>/dev/null)
    
    # If we successfully found an upstream URL, use it. 
    if [ -n "$UPSTREAM_URL" ]; then
        FINAL_URL="$UPSTREAM_URL"
    else
        FINAL_URL="UNKNOWN"
    fi
else
    # It's already a standard web URL (GitHub, GitLab, etc.)
    FINAL_URL="$RAW_URL"
fi

# 3. Output the result
echo "    git_remote_url = **\"$FINAL_URL\";"**
```

The above though should be made a function. It would also be possible to check for CI build specifically in the function by basically checking for a GitHub variable like `$GITHUB_REPOSITORY` in the function. This function will be needed to complete the other ticket.

ShellScripts/Linux/download_github.sh
mcarans: See @LW-archlinux comment on subshells: https://github.com/OoliteProject/oolite/issues/677#issuecomment-4696264858

src/SDL/MyOpenGLView.m
mcarans: If you prefer I can implement that when I look at the other issue https://github.com/OoliteProject/oolite/issues/677

ShellScripts/Linux/install_packages_root.sh
mcarans: This is done

ShellScripts/Linux/download_github.sh
mcarans: This is done

src/SDL/MyOpenGLView.m
mcarans: I've added this function but realise don't actually need it for this. Can just add somethign like thsi to get_version.sh:

```
if [[ -n "$GITHUB_REPOSITORY" ]]; then
    if [[ "$GITHUB_REPOSITORY" == "OoliteProject/oolite" ]]; then
        BUILDER="OoliteProject"
    else
        BUILDER="unknown"
    fi
else
    BUILDER="unknown"
fi
```

mcarans:
I've made changes to resolve a few things. This PR now also fixes https://github.com/OoliteProject/oolite/issues/677

The one thing left is to actually get the version and builder in gitversion.sh in all cases. Builder can be obtained like this I think:

```
if [[ -n "$GITHUB_REPOSITORY" ]]; then
    if [[ "$GITHUB_REPOSITORY" == "OoliteProject/oolite" ]]; then
        BUILDER="OoliteProject"
    else
        BUILDER="unknown"
    fi
else
    BUILDER="unknown"
fi
```

mcarans:
appimage and flatpak built although haven't tested: https://github.com/OoliteProject/oolite/actions/runs/27862543606/job/82460880582

nsis didn't.

mcarans:
Builder is correct when built through CI in appimage. Need to check flatpak and Windows.

.github/workflows/build-all.yaml
mcarans: I've removed common

mcarans:
Flatpak is not correct. Need to fix this:

image

mcarans:
Windows build doesn't install gitversion! That explains why it fails.

mcarans:
@oocube You had mentioned the complexity of the build process. I agree it's a problem and I think I will expand this PR to simplify it. I think it is necessary to get all the builds working reliably and will make maintenance and debugging much easier in future. 

Currently, most builds go through the `ShellScripts/common/build_oolite.sh` script to then call make which uses the Makefile. I will remove both these steps. Instead I will create a `build.sh` script in the root that is the central point for all building and the one place where version and builder are calculated. They will then be passed as parameters everywhere else, so calls to `get_version.sh` will be removed from all other scripts. Other scripts like `create_appimage.sh` will become functions called from that central build script. Those functions will be pushed the version and builder as parameters rather than pulling them in via environment variables from `get_version.sh`. It will be much easier to see what is going on and what is being passed where. 

The version will also be passed to meson as an option instead of calling `get_version.sh`. This is probably needed to fix the flatpak build. It will certainly be much easier to see why version is blank if it is passed from place to place in a parameter. I think the `build.sh` script should have optional parameters `ver_full` and `builder` which if supplied will be used instead of calculating version or  builder. This will support the flatpak build where those parameters can be hardcoded directly into the generated flatpak YAML rather than passing in environment variables.

Can you have a look at https://github.com/OoliteProject/oolite/pull/680? I will need to merge those changes to the meson build into this PR.

oocube:
If the build process can be simplified we ought to go for it. Making massive changes in PRs targeting distinct improvements is not a good way in my eyes. That dilutes the purpose, and we have no chance to accept one change without the other.

Rather let's fix what we tried to fix here and run a separate PR for other improvements.

mcarans:
Flatpak works. Windows packages build - have not tried to run them yet.

Only release job fails.

mcarans:
The version can be parsed from appimage artifact packageinfo in release job. Will look into that tomorrow:

eg.
```
{
        title = "Oolite core";
        identifier = "org.oolite.oolite";

        version = "1.93.1-PullRequest658.181";
        debug_functionality_support = yes;
        required_oolite_version = "1.93";

        license = "GPL 2+ / CC-BY-NC-SA 3.0 - see LICENSE.md for details";
        author = "Giles Williams, Jens Ayton and contributors";
        information_url = "https://oolite.space/";
}

```

mcarans:
On Windows although builder is correct, version has dirty in it unlike appimage and flatpak. Something must be added or changed in the repo during build on Windows only: OoliteInstall-1.93.1-localSemver.11+dirty.0-win-test.exe

oocube:
> On Windows although builder is correct, version has dirty in it unlike appimage and flatpak. Something must be added or changed in the repo during build on Windows only: OoliteInstall-1.93.1-localSemver.11+dirty.0-win-test.exe

The dirty tag is attached if there are uncommitted changes. The counter behind the tag should tell how many uncommitted changes there are. I am wondering why you are getting zero, since zero uncommitted changes should resemble a clean checkout. 

But then the logic does no longer look the way I implemented it. Have a closer look at https://github.com/OoliteProject/oolite/blob/localSemver/ShellScripts/common/get_version.sh#L42

mcarans:
Seems to be a new issue with AnyLinux appimage tool: https://github.com/pkgforge-dev/Anylinux-AppImages/issues/688

mcarans:
Yes it's odd. I have fixed the release so now it's easy to see the versions in the filenames: https://github.com/OoliteProject/oolite/releases/tag/1.93.1-localSemver.11. AppImage is correct which suggests the logic in get_version.sh is ok, but the rest have dirty which is odd. Something must change in the repo during the build for Windows and flatpak only. 

flatpak now seems to miss version which I thought I'd already fixed.

mcarans:
https://github.com/OoliteProject/oolite/releases/tag/1.93.1-localSemver.12 flatpak is now correct (again). Windows still left.

mcarans (dismissed):

mcarans (commented):
Looks better now. It was a Windows line ending issue with git.  https://github.com/OoliteProject/oolite/releases/tag/1.93.1-localSemver.15

phkb:
Is this something to be concerned about?

```
$ make release
meson setup build/meson_release -Ddebug=false -Dstrip_bin=true -Db_lto=true -Dgithub_repository= --native-file clang.ini --reconfigure 2>/dev/null || meson setup build/meson_release -Ddebug=false -Dstrip_bin=true -Db_lto=true -Dgithub_repository= --native-file clang.ini

ERROR: Unknown option: "github_repository".
Directory already configured.

Just run your build command (e.g. ninja) and Meson will regenerate as necessary.
Run "meson setup --reconfigure" to force Meson to regenerate.

If build failures persist, run "meson setup --wipe" to rebuild from scratch
using the same options as passed when configuring the build.

```

mcarans:
> Is this something to be concerned about?
> 
> ```
> $ make release
> meson setup build/meson_release -Ddebug=false -Dstrip_bin=true -Db_lto=true -Dgithub_repository= --native-file clang.ini --reconfigure 2>/dev/null || meson setup build/meson_release -Ddebug=false -Dstrip_bin=true -Db_lto=true -Dgithub_repository= --native-file clang.ini
> 
> ERROR: Unknown option: "github_repository".
> Directory already configured.
> 
> Just run your build command (e.g. ninja) and Meson will regenerate as necessary.
> Run "meson setup --reconfigure" to force Meson to regenerate.
> 
> If build failures persist, run "meson setup --wipe" to rebuild from scratch
> using the same options as passed when configuring the build.
> ```

I'm guessing meson is caching something from an old build from the message "Directory already configured." Try running `make clean`. The github_repository option is defined:
https://github.com/OoliteProject/oolite/blob/localSemver/meson.options#L48-L49

phkb:
image

I'm getting "Oolite v by unknown" as the title on the Oolite window.

mcarans:
> image
> 
> I'm getting "Oolite v by unknown" as the title on the Oolite window.

The unknown is ok - only CI build of OoliteProject repo will show OoliteProject with everything else showing unknown as per oocube's design to distinguish official builds from all else. 

Do you have gitversion installed? `ShellScripts/Windows/install_deps.sh` installs it.

mcarans:
I just added a check for gitversion's existence in gitversion.sh.

phkb:
OK, I ran the install_deps.sh script, and now I get this:

image

Which I assume is what we want, correct?

mcarans:
> OK, I ran the install_deps.sh script, and now I get this:
> image
> 
> Which I assume is what we want, correct?

Yes that's right.

phkb (approved):
Tested all the NSIS installers, everything looks fine to me.

PR #688, opened 2026-06-27
Central mk.sh script

### LINKED ISSUE #671: Make build names (snapshot) match filename identifier (dev)
I find the relationship between builds and identifiers in filenames somewhat baffling.

Currently we have:

1. "release" which is in fact a test package with "test" in the filename
2. "release-deployment" which is the actual release and doesn't add anything to the filename
3. "release-snapshot" which is a development package with "dev" in the filename. 

What I suggest is:

1. "release" which is the actual release and doesn't add anything to the filename
2. "release-test" which adds "test" to the filename
3. "release-dev" which adds "dev" to the filename. 

This would involve updating names and variables etc. that have deployment or snapshot in them to be consistent throughout the project.

What are your thoughts @oocube @phkb @kanthoney ?

### LINKED ISSUE #684: Simplify build: replace Makefile with build.sh, create versions and dates once, pass them by parameter not env var

mcarans:
Versions and build dates all obtained once in meson setup project function
Everything passed by parameter not env var (with necessary exception for Flatpak where host passes env vars in YAML)
Use bash functions rather than scripts where possible
Targets now named deployment, dev, test and debug (no snapshot etc.)
Updated compiler flags to use the same naming for consistency and clarity
Remove Makefile

Fixes https://github.com/OoliteProject/oolite/issues/684
Fixes https://github.com/OoliteProject/oolite/issues/671

See usage of mk.sh in README:  https://github.com/mcarans/oolite/blob/mk_script/README.md
You can type `./mk.sh help` for help on the actions, targets and options


mcarans:
Builds are  here: https://github.com/OoliteProject/oolite/actions/runs/28303563872/attempts/1
(I forgot that release is skipped for PRs and reran the release job unnecessarily)

I've tested on Windows and Linux and looks good to me.

EDIT: builds are here https://github.com/OoliteProject/oolite/actions/runs/28349179837?pr=688

phkb:
I noticed the "Schemata" folder is not being copied into the Resources folder. Which means the verifyOXP option will have difficulty doing anything. I think this has been the case for a little while now, certainly before this PR. I'm happy to create a separate issue to track the resolution, though.

mcarans:
@phkb Good you noticed that. Yes please create a separate issue as it's a bug that precedes this PR. Let me know if you see any other problems.

mcarans:
@phkb @oocube this PR is ready for review

oocube:
I'd rather finish a design discussion on the forum first.

mcarans:
I've set this to draft. While this works well on the command line, I think that there could be problems with IDEs so I'm going to try to move the single version creation into the meson to follow the meson way and be consistent.

mcarans:
This is now ready for review

phkb:
I tried just running 
`./mk.sh build deployment`
but it failed.
I then did
```
./mk.sh clean deployment
./mk.sh build deployment
```
And that did the build. I assume this is because in the first instance I had some previous build in the deployment folder and the script wasn't able to work with it. It might be worth mentioning in the instructions somewhere. (ie "You should always do a 'clean' before doing a 'build'", or whatever makes more sense)

Is the "./" prefix for all the commands necessary? It feels like we're going backwards when the previous iterations were like:
```
make deployment
make debug=no
```

mcarans:
@phkb On clean before build, usually that shouldn't be necessary, but maybe the changes in this PR are large enough that existing build folders won't work. Once the build settles down to something that doesn't change much (hopefully once this PR is merged), running build without clean should be fine.

There are a couple of possibilities for avoiding `./`:

1. Bash alias that can be added to `~/.bashrc` eg.

```
> alias build="./mk.sh build"
mcarans@Holly [~/Code/OoliteRelated/oolite_fixes] ± mk_script → origin {1} ?:11 ✗                                                                                                                                                                                  [py-'python'] [19:53:35]
> build deployment
--> Running Meson setup for: deployment
...
```

Here I have aliased both the command and a specific action, but you may prefer to just alias the command, for example `alias mk="./mk.sh"`

2. Add project folder to path using absolute path for security reasons

EDIT: edited alias to not alias make (which would cause conflict if system make is needed)

mcarans:
@phkb I've updated the README with:

If for some reason the build fails, you can try cleaning the build folder first, although meson should not usually
require this step:

```bash
./mk.sh clean dev
```

oocube:
> Is the "./" prefix for all the commands necessary? It feels like we're going backwards when the previous iterations were like:

Windows and Linux are not that much different. Here is an article that explains:
https://www.howtogeek.com/current-directory-isnt-in-path-why-not-and-should-you-fix-it/

The reason why Linux is more aware of security is simply the history of being a multi user system where DOS/Windows used to be a single user system.

mcarans:
> > Is the "./" prefix for all the commands necessary? It feels like we're going backwards when the previous iterations were like:
> 
> Windows and Linux are not that much different. Here is an article that explains: https://www.howtogeek.com/current-directory-isnt-in-path-why-not-and-should-you-fix-it/
> 
> The reason why Linux is more aware of security is simply the history of being a multi user system where DOS/Windows used to be a single user system.

I think the alias with relative path is the better option as it is specific to the script. That way if you have two project folders, it will run the mk.sh in the current working directory of the specific project. 

On Windows, this could be added to the script that sets up MSYS2 if desired much like how the sourcing of the GNUstep.sh used to be added to .bashrc in that script.

phkb:
I understand this is a difference in the way Linux works (well, I guess it's "bash" really, but still) to how Windows works. Being a long-time Windows user, I'm used to the expectation that, if a app or script is in my current folder, I can simply type the name to execute it. There's just a disconnect between being on a Windows PC, typing a command into a command line, and not getting what I expect. 

I also appreciate that we don't want to keep making things work differently on different OS's. The fewer differences, the easier it is to maintain. 

Do I like having to use "./" in front of the command? No. 
Can I get used to it? Well, you can get used to anything, right? 😃 

The point being, as long as the instructions are clear, I'm good.

mcarans:
@phkb I'm happy to add the alias for Windows if it makes things easier. I can also document how to add it on Linux for consistency. That way you could type `mk` instead of `./mk.sh` for example. (Note that with the alias, you would only be able to type `mk` if the current directory is the project root, whereas you can type the path to `mk.sh` from any folder and the script automatically changes directory to the project root.)  

I would prefer to do that in a separate PR to avoid adding to this already large PR. Please make an issue for it if you'd like that feature (assuming this PR is merged).

oocube:
@mcarans thank you for trying to separate issues

phkb:
No, leave it for now. It's a pretty small issue in the grand scheme of things. There are much bigger fish to fry.

phkb (approved):
I know I'm just a Windows user here, and my needs are "Can I compile a working game?" rather than anything more complex. But this looks fine to me.

oocube (approved):

PR #732, opened 2026-08-30
use arch rather than hardcode x86_64 in espeak-ng-data paths

mcarans:
Basically does what's in this PR:
https://github.com/OoliteProject/oolite/pull/557

### DIFF
--- a/ShellScripts/Linux/install_packages_root.sh
+++ b/ShellScripts/Linux/install_packages_root.sh
@@ -88,9 +88,9 @@ run_script() {
         fi
         install_gitversion "$outputdir"
         local espeak_folder="$(uname -m 2>/dev/null || echo x86_64)-linux-gnu"
-        if [[ ! -d /usr/share/espeak-ng-data &&
-              ! -d /usr/local/share/espeak-ng-data &&
-              ! -d /usr/lib/$espeak_folder/espeak-ng-data ]]; then
+        if [[ ! -d "/usr/share/espeak-ng-data" &&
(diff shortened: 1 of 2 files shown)


phkb (dismissed):

mcarans:
@oocube @LW-archlinux are either of you able to test this?

phkb (approved):

LW-archlinux:
install_packages_root.sh is only called from test_builds.yaml and build-all.yaml which are in .github/worklows folder .
It can only be tested on systems with access to github CI, probably limited to github.com/OoliteProject/oolite and forks within the github ecosystem.
Archlinux espeak-ng package uses /usr/share/espeak-ng-data/ and this is the upstream default path so the change should work for archlinux systems with espeak-ng installed and running on github CI.

As for paths surrounded by double quotes : paths without spaces in them don't need them.
If double-quoting all paths in oolite sourcecode is desired/deemed necessary that should be clarified in a coding style document.
Does oolite have such a document ?

I think post_build.sh is called by meson install so it should affect all builds ?
I'll have to apply this PR to master to test if that poses issues , will try to do that soon.

LW-archlinux:
No issues with the changes on my end.

mcarans:
@LW-archlinux Thanks for testing. On your questions, no there isn't a coding standard. I would like to introduce Clang formatting through the ObjC code and generally to have coding standards everywhere. I'm unsure if there is a wider appetite for that. 

Yes post-build is called as part of the build so Linux builds will fail if espeak data isn't found.

PR #735, opened 2026-09-06
Replace more Win32 code

mcarans:
- isOutputDisplayHDREnabled in MyOpenGLView uses SDL3 instead of Win32
- isRunningOnPrimaryDisplayDevice uses SDL3 instead of Win32
- grabMouseInsideGameWindow uses SDL3 instead of Win32
- SDL_GetDisplayUsableBounds works on Windows too
- Add getDisplayDimensions currently for Windows only although it's SDL3 so would work on Linux
- Should only affect Windows as these functions are in an #if OOLITE_WINDOWS section


phkb (approved):
This looks OK to me, and HDR still works fine as far as I can test it.

PR #740, opened 2026-09-15
Remember last saved game and go to it when loading commander

### LINKED ISSUE #710: Load commander: How about not having Jameson pre-selected?
Dear Developers

I find that good ol' Jameson is pre-selected in the list of commanders that one can load, even though I have a save file the name of which begins with 'aaa_'. When testing OXP and specifically when it is easiest to test with a particular commander, having to move away from Jameson becomes a chore.

Oolite version 1.93.1-201 by OoliteProject (x86-64 test release) under Windows 10.0.26200.8875  [which is actually Windows 11] 64-bit.

EDIT: here is what I see (partially redacted).

Image

mcarans:
Fixes https://github.com/OoliteProject/oolite/issues/710


mcarans:
Although this fixes the behaviour mentioned in the issue (and thus makes the behaviour the same as if there were no Jameson saved game), I'll look into the improvements requested later in the issue.

mcarans:
@LinuxOnTheDesktop confirms this works: https://github.com/OoliteProject/oolite/issues/710#issuecomment-5683564634

mcarans:
@oocube Please can you check

oocube (approved):
I tested Oolite-1.93.1-PullRequest740.3148-dev-x86_64.AppImage and it shows the announced change. I did not encounter problems during that time.

PR #744, opened 2026-09-17
Remove usage of an undocumented Microsoft API

mcarans:
Windows only: Remove usage of an undocumented Microsoft API (looks like it is something to do with dark mode)
Windows only: Grabbing mouse on window move shouldn't be needed with SDL3

Tested on Windows and Linux and works ok for me. I have dark mode on Windows and it looks fine to me without this code.


oocube (approved):
I tested Oolite-1.93.1-PullRequest744.3148-dev-x86_64.AppImage and saw no problems, But then I only tried using the mouse in the menu. During flight I do not touch the mouse at all so I would not notice any difference anyway.
Maybe we get some more qualified feedback?

mcarans:
@phkb does this work for you?

phkb (approved):
Dark/Light mode still works, as does mouse control. Looks good to me.

